PR #1, opened 2016-04-08
Updated DB config to pull prod info from the environment for security

obulkin:


lasley:
:+1: 
^-- that's the sign of a peer review in favor of merge. We should wait for two until actually merging

Suggestions to implement/discuss before merge would be inline. This one is straightforward though

kenortiz:
This looks good to me. I like how you refactored the database.yml. I wasn't sure why you used an environment variable for production at first but after researching it, it makes sense now. Just in case anyone else is curious, I found these articles helpful:
[Heroku Article](https://devcenter.heroku.com/articles/rails-database-connection-behavior)
[Rails Article](http://guides.rubyonrails.org/configuring.html#configuring-a-database)

PR #2, opened 2016-04-09
Updated Gemfile to use the latest Rails version for state-of-the-art …

obulkin:
…goodness


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '4.0.1'
```
lasley, line 4: I'm all for new, but the version should still be defined here. This will force the dependency, so you can guarantee potential users have all of the features you expect to exist in the framework. I recommend a `~>` instead though

lasley, line 4: (note that we do this instead of just the lock so that minor versions can be increased without needing to change things)

lasley:
:+1: pending comments inline

Gemfile
obulkin, line 4: I don't quite follow. As I understand it, the Gemfile.lock should keep all our different deployments in sync: http://stackoverflow.com/questions/6927442/what-is-the-difference-between-gemfile-and-gemfile-lock-in-ruby-on-rails. Am I missing something?

lasley, line 4: You're totally right - I don't think it actually works for Rails though. Take a look at your `.lock` for rails:

`rails`

lasley, line 4: In hindsight, I probably should have led with that lol

obulkin, line 4: I think that portion at the end is just a summary of the explicit dependencies specified in the Gemfile. The longer list at the beginning seems to actually outline all the gems being used in the app and does include a very specific Rails version (line 78 in the new file). Unfortunately, I can't find any detailed docs that would confirm this

lasley, line 4: Hmmm seems like theres been a bunch of debate, a few patches, some more debate, and the community has settled on no versions in Gemfile. TIL

lasley, line 4: I give this :+1: then

obulkin, line 4: Cool. Always good to be on the same page as your community. :)

lasley, line 4: https://mlpforums.com/uploads/monthly_06_2015/post-20300-0-98845900-1435445700.jpg

AidanZeber:
Reviewed pull request, approve of changes and alignment in comments.

obulkin:
Thanks @AidanZeber. I'll delete the branch

PR #4, opened 2016-04-12
Setup OmniAuth

kenortiz:
#### Providers

---

Providers were discussed in Slack on 4/9.
- Facebook
- Google
#### Steps Taken

---
- Installed omniauth-google-oauth2 and omniauth-facebook gems
- Added columns for provider and uid in users table
- Declared providers in devise.rb
- Made User model omniauthable
- Defined omniauth callback routes in routes.rb
- Generated OmniauthCallbacks controller and added method for each provider. If callback is successful, user will be redirected with successful authentication, otherwise user will be redirected to new user registration.
#### Resources

---

Devise Wiki: https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview
Google-oauth2: https://github.com/zquestz/omniauth-google-oauth2


kenortiz:
Fixed syntax error that caused test to fail.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,26 @@
+class OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  def google_oauth2
```
lasley: Repeating yourself in these two methods. Maybe a centralized `do_omniauth` method accepting the kind & session key to append to?

lasley: Also, bravo implementing the two instead of the barebones requirement

test/controllers/omniauth_callbacks_controller_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class OmniauthCallbacksControllerTest < ActionController::TestCase
```
lasley: How do we know this works without tests? :-1: 

https://github.com/intridea/omniauth/wiki/Integration-Testing

lasley:
:-1: - comments inline

lasley:
Also - conflicts exist that will need to be resolved in this branch before a merge to master

config/initializers/devise.rb
```diff
@@ -239,6 +239,8 @@
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
   # config.omniauth :github, 'APP_ID', 'APP_SECRET', scope: 'user,public_repo'
+  config.omniauth :facebook, ENV["FACEBOOK_CLIENT_ID"], ENV["FACEBOOK_CLIENT_SECRET"]
```
obulkin, line 244: What's your approach for setting these environment variables? Codeship can take care of itself, but we should probably use a tool like Figaro to manage this in dev, as well as for manual testing/deployments.

Also, whose keys are you using?

config/routes.rb
```diff
@@ -1,5 +1,5 @@
 MANIACChess::Application.routes.draw do
-  devise_for :users
+  devise_for :users, :controllers => { :omniauth_callbacks => "users/omniauth_callbacks" }
```
obulkin: Right here, you're listing the callback controller as part of the Users namespace, but you don't actually have it set up that way, so I suspect the routing will fail.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    @user = User.from_omniauth(request.env["omniauth.auth"])
+
+    if @user.persisted?
+      flash[:notice] = "devise.omniauth_callbacks.success", :kind => "Google"
```
obulkin: This use of the flash doesn't look right. You're just passing in a string literal with the value "devise.omniauth_callbacks.success", and I don't know how the portion after the comma will be interpreted. I believe the syntax you're looking for is `I18n.t('devise.omniauth_callbacks.success', :kind => 'Facebook')`.

That should pull the right value from the Devise localization file (config/locales/devise.en.yml) using the Rails I18n gem: http://guides.rubyonrails.org/i18n.html

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    if @user.persisted?
+      flash[:notice] = "devise.omniauth_callbacks.success", :kind => "Facebook"
+      sign_in_and_redirect @user, :event => :authentication
+    else
+      session["devise.facebook_data"] = request.env["omniauth.auth"]
+      redirect_to new_user_registration_url
+    end
+  end
+
```
obulkin: You didn't provide the failure method mentioned in the wiki. I'm not sure exactly when this would get triggered, but we should probably include it just in case. I would have it redirect to the root and push an alert into the flash.

app/models/user.rb
```diff
@@ -2,5 +2,13 @@ class User < ActiveRecord::Base
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable and :omniauthable
   devise :database_authenticatable, :registerable,
-         :recoverable, :rememberable, :trackable, :validatable
+         :recoverable, :rememberable, :trackable, :validatable,
+         :omniauthable, :omniauth_providers => [:google_oauth2, :facebook]
+
+  def self.from_omniauth(auth)
+    where(provider: auth.provider, uid: auth.uid).first_or_create do |user|
```
obulkin, line 13: I think we should use `first_or_create!` here so that we're generating a very obvious error if the logic in this method later comes into conflict with some validations.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,26 @@
+class OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  def google_oauth2
+    @user = User.from_omniauth(request.env["omniauth.auth"])
+
+    if @user.persisted?
+      flash[:notice] = "devise.omniauth_callbacks.success", :kind => "Google"
+      sign_in_and_redirect @user, :event => :authentication
+    else
```
obulkin, line 10: I would add an appropriate alert to the flash as part of this fallback scenario. Also, we shouldn't bother with saving the response to the session unless we want to somehow try and use it later on.

obulkin:
:-1: for now. Please see my comments in line. Also, it doesn't look like you've edited any of the Devise views to actually provide users with links that will trigger a sign in via Facebook or Google. I would recommend making that a separate task, since we should ideally add Simple Form and regenerate the Devise views to leverage it before making custom changes to them.

Keep chugging along! :smile: This is a massive task.

app/models/user.rb
obulkin, line 13: Actually, scratch that. If there's a corrupted response coming back from an OmniAuth provider, it could lead to a failure in this method, and we probably want that to happen more gracefully.

config/initializers/devise.rb
kenortiz, line 244: I haven't set up the keys yet. I wanted to configure the basic Omniauth setup asap and I figured we could discuss it either in Slack or in the team meeting on Monday. 

I figured we could use Figaro and set Heroku with the keys, like we have in previous FHP apps, but again I wanted to discuss it with the group first.

config/routes.rb
kenortiz: I was little hesitant to add the users controller since I wasn't sure how we wanted to handle it but I can go ahead and generate a users controller so routing doesn't fail for now.

config/initializers/devise.rb
obulkin, line 244: Got it. Sounds good. We can discuss in chat or during the meeting tomorrow.

app/controllers/omniauth_callbacks_controller.rb
kenortiz: That line was straight from the omniauth-google-oauth2 github page. 
`flash[:notice] = I18n.t "devise.omniauth_callbacks.success", :kind => "Google"`
https://github.com/zquestz/omniauth-google-oauth2

I removed the I18n.t part since I wasn't sure if our app supports Internationalization. From what I understand, I18n.t would be used here if we wanted to provide a translation for the success message. I could be wrong though, I've never used this in past apps. 

I can certainly add it back to that line.

app/controllers/omniauth_callbacks_controller.rb
kenortiz: I wasn't sure when it would be triggered as well since the provider methods redirect to new user registration url, if @user.persisted? is false. Several other resources I checked out, including the devise section in the omniauth-google-oauth2 page, did not have the failure method.

config/routes.rb
obulkin: There's no need to add a Users controller. I was referring to `OmniauthCallbacksController`. With this route config, the app will expect it to be located under app/controllers/users and to be part of the `Users::` namespace. You should either update routes.rb (seems easier) or update the controller and its associated helpers, specs, etc. More info here: http://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing

kenortiz: I just revisited one the resources I used. From what I understand, this line is saying not to use the Devise namespace for the omniauth_callbacks controller. 
https://youtu.be/X6tKAUOMzCs?t=2m30s

kenortiz: I might have it mixed up though lol

app/controllers/omniauth_callbacks_controller.rb
obulkin: The problem is that you removed a reference to a module (`I18n`) and a method call (#t). The original code is passing two arguments to a method. The method returns a string based on the data in devise.en.yml and the arguments, and the string is then added to the flash hash with the key `:notice`. 

Once you remove those parts, you're effectively trying to assign two values to one hash key, which may cause a Ruby exception, and even if it doesn't, the resulting value in the hash will literally be "devise.omniauth_callbacks.success", instead of the nicely worded message stored in devise.en.yml under the keys specified by `devise.omniauth_callbacks.success`.

This module is typically used for translations, but in this particular case, we can just leverage it to reference one of the English messages that Devise has already provided for us. According to the guide, it's included in all Rails apps by default starting with Rails 2.2, and is actually used behind the scenes for a wide range of messages, such as Active Record validation errors.

config/routes.rb
obulkin: Yeah. You're getting thrown off by his comment. I'm not suggesting that you use the Devise namespace, which would amount to the controller class being `Devise::OmniauthCallbacksController`. You just need to make sure that the namespace use in your routing config and in your controller is consistent. The `users/` part in `"users/omniauth_callbacks"` is telling Rails to look for a controller under the users namespace (`Users::OmniauthCallbacksController`) in app/controllers/users, but the rest of your code is set up without a namespace, which in this context is just a way to organize related classes. The easiest fix is to change the route to the following, which is how that video has it set up:

`devise_for :users, :controllers => { :omniauth_callbacks => "omniauth_callbacks" }`

app/controllers/omniauth_callbacks_controller.rb
obulkin: This post suggests that there are certain weird failure scenarios where Devise will try to call this method, so it may be safer to include it: http://stackoverflow.com/questions/10173833/devise-omniauth-failure-how-to-debug-it

config/routes.rb
kenortiz: Ahh, I understand now. I'm still not used to the terminology. 

I was looking at multiple resources and probably added  `=> "users/` by mistake. Thanks for catching that!

obulkin: No worries. The wiki that you shared puts everything under the Users namespace, so that's probably where you grabbed it from.

Gemfile
```diff
@@ -51,9 +51,9 @@ gem 'rails_12factor', group: :production
...
 #include bootstrap
-gem 'bootstrap-sass', '~> 3.3.5.1'
-gem 'sprockets', '2.11.0'
+gem 'bootstrap-sass', '~> 3.3.6'
+gem 'sprockets', '>= 3.2'
 
-# OmniAuth Authentication 
+# OmniAuth Authentication
 gem 'omniauth-google-oauth2'
 gem 'omniauth-facebook'
```
kenortiz, line 76: This commit was to resolve the merge conflict I had. I updated the Gemfile in this branch to reflect the latest Gemfile update in #2. I believe the build failed because Codeship has been setup with RSpec. This is the error in Codeship: 
**rspec-core is not part of the bundle. Add it to Gemfile. (Gem::LoadError)**

I'm going to wait until #5 is merged before adding another commit for this.

obulkin, line 76: Looks like that's the reason. If you get a chance to do so, you could review and merge #5 to speed things up. Based on the comments left by @lasley, this branch will need RSpec tests before it can be merged, so that will kill two birds with one stone.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -3,7 +3,7 @@ def google_oauth2
     @user = User.from_omniauth(request.env["omniauth.auth"])
 
     if @user.persisted?
-      flash[:notice] = "devise.omniauth_callbacks.success", :kind => "Google"
+      flash[:notice] = I18n.t "devise.omniauth_callbacks.success", :kind => "Google"
```
kenortiz: Fixed successful authentication flash notice.

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'rails_helper'
+
+RSpec.describe OmniauthCallbacksController, :type => :controller do
```
kenortiz, line 3: Still need to add some more tests and refactor.

spec/rails_helper.rb
```diff
@@ -52,10 +52,16 @@
...
   config.include FactoryGirl::Syntax::Methods
-end
\ No newline at end of file
+
+  # Turn on "test mode" for OmniAuth
+  OmniAuth.config.test_mode = true
+
+  # Include Devise TestHelpers
+  config.include Devise::TestHelpers, type: :controller
+end
```
kenortiz, line 72: Turned on Omniauth test mode and included Devise TestHelpers.

kenortiz:
Saving my work. Still need to add more tests, refactor, add failure method to omniauth_callbacks_controller and add link.

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        :info => { :email => 'example@gmail.com' }
+        })
+
+      request.env["devise.mapping"] = Devise.mappings[:user]
+      request.env["omniauth.auth"] = OmniAuth.config.mock_auth[:facebook]
+      get :facebook
+    end
+
+    it "should successfully authenticate from Facebook" do
+      user = User.first
```
obulkin: Instead of relying on `User.first`, which seems really dependent on a particular DB state, I would use `User.find_by(email: "example@gmail.com")` and just make sure that it's not returning `nil`. For similar reasons of robustness, I would also use different uid and email values in each of the mock auth hashes.

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      expect(user.email).to eq('example@gmail.com')
+    end
+
+    it "should successfully set flash notice" do
+      user = User.first
```
obulkin: There's no reason to retrieve a user instance if you're just checking the flash.

obulkin:
Getting there. 😄  I left a couple small comments in line. Since we're using RSpec, you can just delete the test folder entirely. It contains files generated by the default test framework.

spec/controllers/omniauth_callbacks_controller_spec.rb
kenortiz: Thanks @obulkin, I'll give that a try and good point about the different email and uid values.

kenortiz:
Thanks @obulkin. I deleted the test folders in my local repository but it hasn't updated here since I just committed the files I was working on last night.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -1,26 +1,28 @@
...
       redirect_to new_user_registration_url
     end
   end
 
+  alias_method :facebook, :do_omniauth
+  alias_method :google_oauth2, :do_omniauth
+
+  def failure
+    flash[:alert] = "Could not authenticate."
+    redirect_to root_path
```
kenortiz, line 21: We do not have a root setup yet but since we eventually will, I added this in here.

obulkin, line 21: You'll probably need to set one up in order to test this method. A static pages controller and an empty homepage view should do it.

app/views/devise/registrations/new.html.erb
```diff
@@ -14,7 +14,4 @@
   </div>
 <% end %>
 
-<%= link_to "Sign up with Facebook", user_omniauth_authorize_path(:facebook) %>
-<%= link_to "Sign up with Google", user_omniauth_authorize_path(:google_oauth2) %>
-
 <%= render "devise/shared/links" %>
```
kenortiz, line 17: Adding links in devise views isn't necessary unless we want custom links such as Google or Facebook icons. 
Devise shared links already provides default links:

``` ruby
<%- if devise_mapping.omniauthable? %>
  <%- resource_class.omniauth_providers.each do |provider| %>
    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider) %><br />
  <% end -%>
<% end -%>
```

obulkin, line 17: Good catch. I totally forgot about this. 😞

spec/factories/omniauth.rb
```diff
@@ -0,0 +1,37 @@
+FactoryGirl.define do
+
+  factory :auth_hash, class: OmniAuth::AuthHash do
```
kenortiz, line 5: I added an omniauth auth hash factory just in case future tests are needed.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    if @user.provider == "google_oauth2"
+      provider_kind = "Google"
+    end
+
+    if @user.persisted?
+      flash[:notice] = I18n.t("devise.omniauth_callbacks.success", kind: "#{provider_kind}")
+      sign_in_and_redirect @user, :event => :authentication
+    else
+      session["devise.#{provider_data}_data"] = request.env["omniauth.auth"]
```
obulkin: Is this data actually being used during the standard Devise registration process in some way? If we're not using it and Devise isn't using, we probably shouldn't store it.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      redirect_to new_user_registration_url
+    end
+  end
+
+  alias_method :facebook, :do_omniauth
+  alias_method :google_oauth2, :do_omniauth
+
+  def failure
+    flash[:alert] = "Could not authenticate."
```
obulkin: Could you make this message a bit more user-friendly? I'm thinking something along the lines of "Authentication failed. Please try again later or sign up with MANIAC Chess directly by providing an email and password."

spec/factories/omniauth.rb
```diff
@@ -0,0 +1,37 @@
...
+    skip_create
```
obulkin: FYI, you can also keep factory_girl from saving the object by using `FactoryGirl.build` rather than `FactoryGirl.create` to generate it. `skip_create` is a nice safeguard though.

spec/factories/omniauth.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    transient do
+      sequence(:uid)
+      provider "provider"
+      email { "email" }
+    end
```
obulkin: As far as I understand, transient attributes are intended for situations where the factory needs to receive options that will be used during the object creation process but won't actually be properties of the created object. In this case, since you're just trying to set up some defaults, you can simply list them as normal attributes of the base `auth_hash` factory.

Also, the `attribute {#some logic here}` syntax is used to create lazy attributes whose value is determined at the time of object creation based on the code in the block. This is useful for sequences or situations like setting an attribute equal to the current time, but it's not necessary when you're just setting an attribute equal to a static value like `"email"`.

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        OmniAuth.config.mock_auth[:do_omniauth] = FactoryGirl.create(:auth_hash, :facebook)
+
+        request.env["devise.mapping"] = Devise.mappings[:user]
+        request.env["omniauth.auth"] = OmniAuth.config.mock_auth[:do_omniauth]
+      end
+
+      it "should successfully authenticate using Facebook authentication hash and display a success message in flash" do
+        get :facebook
+        user = User.find_by(email: 'testuser@facebook.com')
+        expect(user.id).not_to eq(nil)
```
obulkin: `user.id` can never be `nil`. Either the user will be in the DB and will have a valid ID, or the user won't be present, and `find_by` will return `nil`.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    if @user.provider == "google_oauth2"
+      provider_kind = "Google"
+    end
+
+    if @user.persisted?
+      flash[:notice] = I18n.t("devise.omniauth_callbacks.success", kind: "#{provider_kind}")
+      sign_in_and_redirect @user, :event => :authentication
+    else
+      session["devise.#{provider_data}_data"] = request.env["omniauth.auth"]
+      redirect_to new_user_registration_url
```
obulkin, line 12: Since this scenario represents some sort of failure as well, we should probably add a user-friendly alert to the flash before redirecting.

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+    end
+
+    context "given user does not persist" do
+      before do
+        OmniAuth.config.mock_auth[:do_omniauth] = FactoryGirl.create(:auth_hash, :facebook, :does_not_persist)
+
+        request.env["devise.mapping"] = Devise.mappings[:user]
+        request.env["omniauth.auth"] = OmniAuth.config.mock_auth[:do_omniauth]
+      end
```
obulkin: Since these `before` blocks each apply to only a single example, they don't provide any benefits. I would clean this up as follows:
- Move `request.env["devise.mapping"] = Devise.mappings[:user]` into a single `before` block at the top of the spec. This is the only line that actually applies to every test.
- Move the other two lines into the `it` block and combine them into one: `request.env["omniauth.auth"] = FactoryGirl.create(:auth_hash, :facebook, :does_not_persist)`

obulkin:
This is looking much better but isn't quite there yet. I left a number of smaller comments in line, which should all be pretty easy to address.

Keep chugging along! 😄

spec/controllers/omniauth_callbacks_controller_spec.rb
kenortiz: Tried that and it didn't work. Did it work on your end?

obulkin: Yup. I just tried it out, and it works fine. The call to `before` should come right after the `RSpec.describe` line.

spec/factories/omniauth.rb
kenortiz: That makes sense.

I was trying to use unique credentials for each test that uses this hash and for some reason(probably saw an example somewhere) used `transient` for `sequence(:uid)` since I couldn't use that in my hash. It works fine in my traits so I removed the `transient` section.

Thanks @obulkin

spec/controllers/omniauth_callbacks_controller_spec.rb
kenortiz: You're absolutely right. I tried it again and it worked, which makes total sense. Not sure what I did wrong when I tried it last night. I probably commented out the wrong line of code or something. Who knows, it was pretty late 😴

obulkin: 😞  Yup. There definitely comes a point where you're so tired that you're breaking two things for each one that you fix.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,24 @@
+class OmniauthCallbacksController < Devise::OmniauthCallbacksController
+
+  def do_omniauth
+    @user = User.from_omniauth(request.env["omniauth.auth"])
+    provider_data = @user.provider
```
obulkin: Optional: This variable isn't being used anywhere else in the code, so you can delete the line.

spec/factories/omniauth.rb
```diff
@@ -0,0 +1,33 @@
...
+    skip_create
+
+    initialize_with do
+      OmniAuth::AuthHash.new({
+        provider: provider,
+        uid: uid,
+        info: {
+          email: email
+        }
+      })
```
obulkin, line 14: This factory will now generate exceptions if it's called without providing a trait because `provider`, `uid`, and `email` will be undefined. That seems too brittle, so could you please add some non-transient, default values? I'm sorry if my comment about transient attributes earlier introduced any confusion.

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+RSpec.describe OmniauthCallbacksController, :type => :controller do
+  before(:each) do
+    request.env["devise.mapping"] = Devise.mappings[:user]
+  end
+
+  describe "#facebook" do
+    it "should successfully authenticate using Facebook authentication hash and display a success message in flash" do
+      request.env["omniauth.auth"] = FactoryGirl.create(:auth_hash, :facebook)
+      get :facebook
+      user = User.find_by(email: 'testuser@facebook.com')
```
obulkin, line 12: You're no longer testing whether a user was properly added to the DB, which is probably something we want to verify. My comment earlier wasn't meant to imply that you should scrap that line but just that it needs to be rewritten. Since `find_by` will return `nil` when it doesn't find a match, `expect(user).not_to eq(nil)` should work here.

spec/factories/omniauth.rb
kenortiz, line 14: I understand that, but that is why I added traits. My intent was to always use a trait in my tests.

app/controllers/omniauth_callbacks_controller.rb
kenortiz: Good catch. I'll delete that.

spec/factories/omniauth.rb
kenortiz, line 14: I'll add a comment to the factory to use a trait when calling this.

obulkin, line 14: Fair enough. I suppose it's reasonable to expect users of a factory to understand how it's been configured before calling it.

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect(user).not_to eq(nil)
+      expect(flash[:notice]).to eq("Successfully authenticated from Facebook account.")
+    end
+
+    it "should redirect the user to the new user registration form" do
```
obulkin: This test description is missing the `if the user does not persist` portion.

kenortiz: Good catch, I must have missed that when I removed context sections and refactored. I'll update that now.

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    request.env["devise.mapping"] = Devise.mappings[:user]
+  end
+
+  describe "#facebook" do
+    it "should successfully authenticate using Facebook authentication hash and display a success message in flash, if the user persists" do
+      request.env["omniauth.auth"] = FactoryGirl.create(:auth_hash, :facebook)
+      get :facebook
+      user = User.find_by(email: 'testuser@facebook.com')
+      expect(user).not_to eq(nil)
+      expect(flash[:notice]).to eq("Successfully authenticated from Facebook account.")
```
obulkin, line 14: I think we should also make sure that the user actually gets signed in during the successful authentication tests. This is pretty easy to do via `subject` (http://www.relishapp.com/rspec/rspec-core/v/3-4/docs/subject/implicitly-defined-subject) and the `current_user` helper provided by Devise.

obulkin:
Woo! 👍  I think this is good to go now. Way to persevere, @kenortiz.

It might be nice to have unit tests for the `from_omniauth` method, but it is being implicitly tested as part of the controller spec, so I'm also fine with leaving these out. It's your call.

kenortiz:
@obulkin I'm fine leaving it out. That method is unchanged from Devise's omniauth setup instructions, except for `user.image` and `user.name`, which I removed since we currently do have those fields. Also, my other tests would have failed if `from_omniauth` wasn't working properly.

obulkin:
@kenortiz -- Sounds good. I'll merge.

As a side note, I feel like the main value of the tests is not in determining whether your code works currently (this can usually be established manually) but in catching future changes that break it accidentally (regressions). As a result, I would say that it doesn't really matter whether the code was taken from a reliable source or written from scratch.

lasley:
Excellent job here @kenortiz - OmniAuth is a rough one.

kenortiz:
Thanks @lasley 😄

PR #9, opened 2016-04-15
Add game model

obulkin:
- Generated a game model
- Built out and ran the associated migration
- Added validations to the model
- Created a base factory for the model
- Wrote tests for the model logic and any associated DB constraints

### DIFF
--- a/app/models/game.rb
+++ b/app/models/game.rb
@@ -1,0 +1,2 @@
+class Game < ActiveRecord::Base
+end
--- a/app/models/game.rb
+++ b/app/models/game.rb
@@ -1,2 +1,6 @@
 class Game < ActiveRecord::Base
+  validates :name, presence: true
+  validates :white_player_id, presence: true
+  validates :black_player_id, presence: true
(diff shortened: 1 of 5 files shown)


app/models/game.rb
```diff
@@ -0,0 +1,12 @@
+class Game < ActiveRecord::Base
+  validates :name, presence: true
+  validates :white_player_id, presence: true
+  validates :black_player_id, presence: true
```
lasley: We will run into issues down the road with these being required due to needing to join a game. There's not going to be many instances in which a game will be created immediately with both players. Maybe just a required on whichever the first player is?

spec/models/game_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'rails_helper'
+
+RSpec.describe Game, type: :model do
```
lasley, line 3: Excellent tests!

app/models/game.rb
obulkin: Good point. I'll replace these with a custom validation that looks for at least one ID and will remove the constraints on the columns.

lasley: :+1:

PR #10, opened 2016-04-16
Add piece model

obulkin:
- Generated a piece model
- Built out and ran the associated migration
- Added validations to the model
- Created a base factory for the model
- Wrote tests for the model logic and any associated DB logic

### DIFF
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -1,0 +1,2 @@
+class Piece < ActiveRecord::Base
+end
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -1,2 +1,7 @@
 class Piece < ActiveRecord::Base
+  validates :type, presence: true, inclusion: {within: ["Pawn", "Rook", "Bishop", "Knight", "Queen", "King"]}
+  validates :color, presence: true, inclusion: {within: ["white", "black"]}
+  validates :rank, presence: true, inclusion: {within: 1..8}
+  validates :file, presence: true, inclusion: {within: 1..8}
+  validates :game_id, presence: true
 end
--- a/db/migrate/20160414225123_create_pieces.rb
+++ b/db/migrate/20160414225123_create_pieces.rb
@@ -1,0 +1,16 @@
+class CreatePieces < ActiveRecord::Migration
+  def change
+    create_table :pieces do |t|
+      t.string :type, null: false
+      t.string :color, null: false
+      t.integer :rank, null: false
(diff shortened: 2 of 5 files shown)


app/models/piece.rb
```diff
@@ -0,0 +1,7 @@
+class Piece < ActiveRecord::Base
```
lasley, line 1: Take a look at scopes - http://katieleonard.ca/blog/2015/scopes-and-single-table-inheritance/

obulkin, line 1: Is there a specific spot where you think we should be using these? A scope for active games would be good, since this involves some `OR` logic, but I don't want to go overboard with setting up scopes for situations that can easily be handled with `where(property: value)`.

Btw, I'm planning to set up all the connections between the three tables as part of a separate ticket: https://trello.com/c/Srt3LB1I/32-create-connections-between-the-user-game-and-piece-models

lasley, line 1: Should only use it for the STI IMO. Basically you're breaking string type definitions into contextual symbols.

Cool, I'm all for isolated changesets.

obulkin, line 1: I don't follow. My understanding is that Rails should take care of all the STI scoping automatically, as long as the models inherit properly from Piece and have class names that match up with the values in the `type` column. That's why they're all capitalized. 

I tested it in dev, and it seemed to work fine.

lasley, line 1: Cool works for me, I'm probably just out of date on that :+1:

obulkin, line 1: Or are you talking about scoped associations on the game model? I feel like `@game.pieces.where(type: "Rook", color: "white")` is pretty straightforward.

obulkin, line 1: Actually, I think I was confused earlier. Rails doesn't automatically give you support for something like `Piece.pawns`, but I feel like there are so many alternatives (`Pawn.all`, `Piece.find_by(type: "pawn")`, `Piece.where(type: "pawn")`) that the scopes are overkill. I could ask for a team vote though.

lasley, line 1: Hah I didn't think that was a thing, but I never doubt when Rails adds more automatic things. Seemed weird for something as explicit as scoping to be automated though.

We'll be fine - no need for a vote or to bring back a merged PR.

obulkin, line 1: Sounds good. We could always add this as part of the STI ticket, so there's still a good opportunity, but I'm also fine making do without.

PR #11, opened 2016-04-17
populate game

brandoncha:
-Created a Method to fill a game with pieces that will be triggered when a new game is created
-Hooked up x,y coordinates
-Created models of each piece
-Created association for piece and game models

### DIFF
--- a/spec/models/game_spec.rb
+++ b/spec/models/game_spec.rb
@@ -43,4 +43,9 @@ RSpec.describe Game, type: :model do
       expect(game.white_to_move).to eq(true)
     end
   end
+
+  it "should have correct number of pieces" do
+      game = Game.create
+      game.pieces.count==32
+    end
 end
--- a/app/models/bishop.rb
+++ b/app/models/bishop.rb
@@ -1,0 +1,2 @@
+class Bishop < Piece
+end
--- a/app/models/game.rb
+++ b/app/models/game.rb
@@ -3,6 +3,10 @@ class Game < ActiveRecord::Base
   validates :state, presence: true, inclusion: {within: ["open", "white_in_check", "black_in_check", "white_in_mate", "black_in_mate", "draw"]}
   validate :at_least_one_player, :players_must_be_different
 
+  has_many :pieces
+
+  after_create :populate_board!
+
   private
   def at_least_one_player
     errors.add(:base, "A game must have at least one player") unless white_player_id.present? || black_player_id.present?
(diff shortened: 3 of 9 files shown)


obulkin:
The core code looks good to me, but this needs more specs. Instead of checking for a total piece count, I would test that there is an appropriate piece on each square of the board (e.g. test that there's a white pawn on a2, test that there's a black knight on g8, etc.), which addresses the piece count portion implicitly and also confirms color/position. Since you added two associations, I would create tests for those as well.

For extra spec clarity and DRYness, you should use `describe` blocks to group logically related examples (I did this with the validation and DB constraint tests) and rely on factory_girl to handle object creation for you (see my inline comments for more info).

You're almost there! 😄

spec/models/game_spec.rb
```diff
@@ -43,4 +43,9 @@
       expect(game.white_to_move).to eq(true)
     end
   end
+
+  it "should have correct number of pieces" do
+      game = Game.create
```
obulkin, line 48: This line is not actually writing to the DB due to the validations on `Game`. Instead, you should just use the `game` variable created by the `let` line at the top (more info at http://www.relishapp.com/rspec/rspec-core/v/3-4/docs/helper-methods/let-and-let) or call the factory_girl factory set up in spec/factories/games.rb by using `create :game`.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,9 @@
...
+      game.pieces.count==32
```
obulkin, line 49: The standard approach for something like this is to use an RSpec matcher (https://www.relishapp.com/rspec/rspec-expectations/v/3-4/docs/built-in-matchers), giving you `expect(game.pieces.count).to eq(32)`. Your current setup isn't actually generating a valid test and won't fail even if the count is different. You can see this by changing the line to something like `game.pieces.count == 2` and running RSpec again.

brandoncha, line 49: I see. But I tried to use 'expect(game.pieces.count).to eq(32)' first. And it got 0 not 32.

obulkin, line 49: That was probably the test legitimately failing. It may have been because the line just before it doesn't actually create a valid game and save it to the DB or for some other reason.

brandoncha:
ok, I will make more specs for an appropriate piece on each square of the board and between piece and game. 
Thank you for review.

PR #12, opened 2016-04-18
Initialize board

brandoncha:
I added more specs for correct positions and database relationship between pieces and games.
I tried to make a test like this(ex. test that there's a black knight on g8). But it keeps failed.
Please let me know any suggestions if it should be necessary.

### DIFF
--- a/spec/models/game_spec.rb
+++ b/spec/models/game_spec.rb
@@ -81,8 +81,8 @@ RSpec.describe Game, type: :model do
 
       it "should have a black piece at rank 8, file #{file}" do
         @black_piece = game.pieces.find{|piece| piece.rank == 8 && piece.file == file}
-        @black_piece_type = @black_piece.type
-        expect(@black_piece.type).to eq(@black_piece_type)
+        expected = @black_piece.type
+        expect(@black_piece.type).to eq(expected)
       end
     end  
   end
--- a/spec/models/game_spec.rb
+++ b/spec/models/game_spec.rb
@@ -49,40 +49,33 @@ RSpec.describe Game, type: :model do
       expect(game.pieces.count).to eq 32
     end
 
-    it "should successfully hook up the database relationship between pieces and games" do
-      piece = game.pieces.last
-      expect(piece.game).to eq(game)
-    end
-
     it "should be empty on the squares" do
       expect(game.pieces.where(rank: 3..6)).to eq([])
     end
 
     (1..8).each do |file|
       it "should have a white pawn at rank 2, file #{file}" do
-        @white_pawn = game.pieces.find{|piece| piece.rank == 2 && piece.file == file}
+        white_pawn = game.pieces.find{|piece| piece.rank == 2 && piece.file == file}
         expected = ["Pawn", "white"]
-        expect([@white_pawn.type, @white_pawn.color]).to eq(expected)
+        expect([white_pawn.type, white_pawn.color]).to eq(expected)
       end
 
       it "should have a black pawn at rank 7, file #{file}" do
-        @black_pawn = game.pieces.find{|piece| piece.rank == 7 && piece.file == file}
+        black_pawn = game.pieces.find{|piece| piece.rank == 7 && piece.file == file}
         expected = ["Pawn", "black"]
-        expect([@black_pawn.type, @black_pawn.color]).to eq(expected)
+        expect([black_pawn.type, black_pawn.color]).to eq(expected)
       end
     end
 
     {1 => "Rook", 2 => "Knight", 3 => "Bishop", 4 =>"Queen", 5 => "King", 6=> "Bishop", 7 => "Knight", 8 => "Rook"}.each do |file, type|  
-      it "should have a white piece at rank 1, file #{file}" do
-        @white_piece = game.pieces.find{|piece| piece.rank == 1 && piece.file == file}
-        expected = @white_piece.type
-        expect(@white_piece.type).to eq(expected)
+      it "should have a correct type of white piece at rank 1, file #{file}" do
+        white_piece = game.pieces.find{|piece| piece.rank == 1 && piece.file == file}
+        expect(white_piece.type).to eq(type)
       end
 
-      it "should have a black piece at rank 8, file #{file}" do
-        @black_piece = game.pieces.find{|piece| piece.rank == 8 && piece.file == file}
-        expected = @black_piece.type
-        expect(@black_piece.type).to eq(expected)
+      it "should have a correct type of black piece at rank 8, file #{file}" do
+        black_piece = game.pieces.find{|piece| piece.rank == 8 && piece.file == file}
+        expect(black_piece.type).to eq(type)
       end
     end  
   end
--- a/spec/models/game_spec.rb
+++ b/spec/models/game_spec.rb
@@ -75,8 +75,8 @@ RSpec.describe Game, type: :model do
     {1 => "Rook", 2 => "Knight", 3 => "Bishop", 4 =>"Queen", 5 => "King", 6=> "Bishop", 7 => "Knight", 8 => "Rook"}.each do |file, type|  
       it "should have a white piece at rank 1, file #{file}" do
         @white_piece = game.pieces.find{|piece| piece.rank == 1 && piece.file == file}
-        @white_piece_type = @white_piece.type
-        expect(@white_piece.type).to eq(@white_piece_type)
+        expected = @white_piece.type
+        expect(@white_piece.type).to eq(expected)
       end
 
       it "should have a black piece at rank 8, file #{file}" do
--- a/spec/models/game_spec.rb
+++ b/spec/models/game_spec.rb
@@ -43,4 +43,9 @@ RSpec.describe Game, type: :model do
       expect(game.white_to_move).to eq(true)
     end
   end
+
+  it "should have correct number of pieces" do
+      game = Game.create
+      game.pieces.count==32
+    end
 end
--- a/spec/models/game_spec.rb
+++ b/spec/models/game_spec.rb
@@ -46,7 +46,7 @@ RSpec.describe Game, type: :model do
 
   describe "Populate Board" do
     it "should have correct number of pieces" do
-      expect(game.pieces.count).to eq 32
+      expect(game.pieces.count).to eq (32)
     end
 
     it "should be empty on the squares" do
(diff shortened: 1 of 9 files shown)


obulkin:
I would approach this as follows:
1. Look for a piece in a specific spot using #where: `game.pieces.where(file: 1, rank: 1)`
2. Save the result to a variable
3. Make sure the result is not just an empty relation
4. Verify the properties of the model object that was returned (color and type)

You can check all the squares that should be empty in one go by using ranges: `game.pieces.where(file: 3..6, rank: 3..6)`.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,33 @@
       expect(game.white_to_move).to eq(true)
     end
   end
+
+  describe "Populate Board" do
+    it "should have correct number of pieces" do
+      game = FactoryGirl.create(:game)
```
obulkin: There's no reason to create a new game for each test. You can simply do this once at the top of the `describe` block and then reference it in all the examples. Also, I've added an RSpec config option that simplifies factory_girl syntax, so you can use `create :game` in place of `FactoryGirl.create :game`.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,33 @@
...
+      expect(game.pieces.count).to eq 32
+    end
+
+    it "Pawn is in correct position" do
+      game = FactoryGirl.create(:game)
+      expect(game.pieces.first.type).to eq("Pawn")
```
obulkin: The order that the board gets populated in doesn't actually matter, so I wouldn't create tests that rely on a specific ordering with syntax like `pieces.first`

spec/models/game_spec.rb
```diff
@@ -43,4 +43,33 @@
...
+      expected = [2, 1]
+      expect([game.pieces.first.rank, game.pieces.first.file]).to eq(expected)
+    end
+
+    it "Rook is in correct position" do
+      game = FactoryGirl.create(:game)
+      expect(game.pieces.last.type).to eq("Rook")
+
+      expected = [8, 8]
+      expect([game.pieces.last.rank, game.pieces.last.file]).to eq(expected)
```
obulkin: This is more of a style preference, but I think it's better to split out multiple comparisons into multiple `expect` statements. That way, they will get executed one at a time, and it will be easier to tell from the test results which one caused the test to fail.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,33 @@
...
+    end
+
+    it "should successfully hook up the database relationship between pieces and games" do
```
obulkin: Since the associations are declared separately and don't have to be bidirectional, we should have a separate test for each one.

brandoncha:
Oleg, thanks for comments.
I will try to fix it and make the test for the pieces position.

obulkin:
Sure thing. BTW, it occurs to me that you can DRY these tests up significantly by using ranges and hashes to build them iteratively. For example, you can do the following for all the white pawn squares:

```
(1..8).each do |file|
  it "should have a white pawn at rank 2, file #{file}" do
    # test stuff
  end
end
```

and something like this for a rank with non-pawn pieces:

```
{1 => "Rook", 2 => "Knight" }.each do |file, type|
  it "should have a white #{type} at rank 1, file #{file}" do
    # test stuff
  end 
end

```

spec/models/game_spec.rb
```diff
@@ -43,4 +43,47 @@
       expect(game.white_to_move).to eq(true)
     end
   end
+
+  describe "Populate Board" do
+    it "should have correct number of pieces" do
+      expect(game.pieces.count).to eq 32
+    end
+
+    it "should successfully hook up the database relationship between pieces and games" do
```
obulkin: This isn't actually part of `populate_board!`, so it shouldn't be tested in this `describe` block. In fact, you can drop this test entirely, since I'll be adding specs for these associations as part of https://trello.com/c/Srt3LB1I/32-create-connections-between-the-user-game-and-piece-models, which is almost ready.

brandoncha: Yes, I also doubted it. It should be in create-connection-between-models. will drop this test.
thanks.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,47 @@
...
+      expect(piece.game).to eq(game)
+    end
+
+    it "should be empty on the squares" do
+      expect(game.pieces.where(rank: 3..6)).to eq([])
+    end
+
+    (1..8).each do |file|
+      it "should have a white pawn at rank 2, file #{file}" do
+        @white_pawn = game.pieces.find{|piece| piece.rank == 2 && piece.file == file}
```
obulkin: An instance variable declared in an `it` block will actually be available across the entire `describe` block that the specific example belongs to and can cause tests to interfere with each other. Just use local variables instead.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,47 @@
...
+        expected = ["Pawn", "black"]
+        expect([@black_pawn.type, @black_pawn.color]).to eq(expected)
+      end
+    end
+
+    {1 => "Rook", 2 => "Knight", 3 => "Bishop", 4 =>"Queen", 5 => "King", 6=> "Bishop", 7 => "Knight", 8 => "Rook"}.each do |file, type|  
+      it "should have a white piece at rank 1, file #{file}" do
+        @white_piece = game.pieces.find{|piece| piece.rank == 1 && piece.file == file}
+        expected = @white_piece.type
+        expect(@white_piece.type).to eq(expected)
```
obulkin: Right here, you're basically just testing for `@white_piece.type == @white_piece.type`, which is not what you want. Instead, you should be comparing against the values in the hash using `type`.

This is why you should probably stop using the following pattern:

```
expected = some_value_or_values
expect(test_result).to eq(expected)
```

It adds an unnecessary step/variable and makes it harder to see what the test is doing. Instead, just use `expect(test_result).to eq(expected_value_or_values)`.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,47 @@
...
+      end
+
+      it "should have a black piece at rank 8, file #{file}" do
```
obulkin: I would make the test description more clear by including `type` in the string.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,40 @@
...
+      expect(game.pieces.count).to eq 32
+    end
+
+    it "should be empty on the squares" do
+      expect(game.pieces.where(rank: 3..6)).to eq([])
+    end
+
+    (1..8).each do |file|
+      it "should have a white pawn at rank 2, file #{file}" do
+        white_pawn = game.pieces.find{|piece| piece.rank == 2 && piece.file == file}
```
obulkin: Instead of using `find` with a block, you should use `where` or `find_by`. In the context of this test, the results may appear to be the same, but behind the scenes the first approach will bring back the entire table from the DB and then evaluate each result in memory, while `where` or `find_by` will incorporate your constraints into the resulting SQL query and bring back only the rows that you need. In a production app with a lot of data, bringing the entire pieces table into memory may be prohibitively resource intensive, so it's not great to get in the habit of doing this.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,40 @@
...
+        expected = ["Pawn", "black"]
+        expect([black_pawn.type, black_pawn.color]).to eq(expected)
+      end
+    end
+
+    {1 => "Rook", 2 => "Knight", 3 => "Bishop", 4 =>"Queen", 5 => "King", 6=> "Bishop", 7 => "Knight", 8 => "Rook"}.each do |file, type|  
+      it "should have a correct type of white piece at rank 1, file #{file}" do
+        white_piece = game.pieces.find{|piece| piece.rank == 1 && piece.file == file}
+        expect(white_piece.type).to eq(type)
+      end
```
obulkin, line 75: You should also check that the piece is the right color, and before you check for either property, I would make sure that the square isn't completely empty. This applies to all the examples that expect to find a piece.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,50 @@
...
+        expected = ["Pawn", "white"]
+        expect([white_pawn.type, white_pawn.color]).to eq(expected)
+      end
+
+      it "should have a black pawn at rank 7, file #{file}" do
+        black_pawn = game.pieces.find_by(rank: 7 , file: file)
+        expect(black_pawn.present?).to eq(true)
+        
+        expected = ["Pawn", "black"]
+        expect([black_pawn.type, black_pawn.color]).to eq(expected)
```
obulkin: You're still using the problematic pattern that I mentioned here: https://github.com/fp-team-maniac/maniac-chess/pull/12#discussion_r60299653. Is this intentional or did you miss that part of the comment? Since this is a style point, I don't want to hold up the merge because of it, but I'd like to understand your reasoning.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,50 @@
...
+      end
+    end
+
+    {1 => "Rook", 2 => "Knight", 3 => "Bishop", 4 =>"Queen", 5 => "King", 6=> "Bishop", 7 => "Knight", 8 => "Rook"}.each do |file, type|  
+      it "should have a correct type of white piece at rank 1, file #{file}" do
```
obulkin: Could you please use `#{type}` in the test description to make it more clear?

spec/models/game_spec.rb
```diff
@@ -43,4 +43,50 @@
...
+      expect(game.pieces.count).to eq 32
+    end
+
+    it "should be empty on the squares" do
+      expect(game.pieces.where(rank: 3..6)).to eq([])
+    end
+
+    (1..8).each do |file|
+      it "should have a white pawn at rank 2, file #{file}" do
+        white_pawn = game.pieces.find_by(rank: 2 , file: file)
```
obulkin: I know I totally suggested using `find_by` here, but it now occurs to me that `where` is a better approach because you can then check the count of the result set to make sure that there aren't multiple pieces on the square in question. `find_by` only returns the first result that matches the given conditions, so it'll mask this scenario. Here's an example of the new approach:

```
pieces = game.pieces.where(rank: 2, file: file)
expect(pieces.count).to eq(1)
# implement other tests using pieces.first
```

Sorry about not mentioning this earlier. Writing thorough tests is pretty tough. There are so many scenarios to consider.

spec/models/game_spec.rb
```diff
@@ -43,4 +43,46 @@
...
+        pieces = game.pieces.where(rank: 2 , file: file)
+        expect(pieces.present?).to eq(true)
+        expect(pieces.count).to eq(1)
+        expect([pieces.first.type, pieces.first.color]).to eq(["Pawn", "white"])
+      end
+
+      it "should have a black pawn at rank 7, file #{file}" do
+        pieces = game.pieces.where(rank: 7 , file: file)
+        expect(pieces.present?).to eq(true)
+        expect(pieces.count).to eq(1)
```
obulkin, line 65: These two lines are redundant. Unlike `find_by`, which returns `nil` when it doesn't find anything, `where` will always return an `ActiveRecord` object that responds to #count, so it should be enough to just verify the count.

obulkin:
Nice. 😄  This looks almost ready to me. I just left one minor comment in line.

spec/models/game_spec.rb
brandoncha, line 65: Ok. i also was doubting if "expect(pieces.present?).to eq(true)" is necessary here.
I will modify it.

obulkin, line 65: Sounds good. One thing I find helpful when I'm in doubt about an implementation detail like this is going to the Rails console and playing around with the methods.

lasley, line 65: And when things get real rough, set a break point - http://guides.rubyonrails.org/v2.3.11/debugging_rails_applications.html#debugging-with-ruby-debug

brandoncha, line 65: I got it. Thanks for suggestions @obulkin @lasley .

brandoncha:
I removed the test for "shouldn't be empty test".

obulkin:
Awesome! It took a few revisions, but this now looks good to go. 👍

brandoncha:
Finanlly!!
@obulkin Thank you for help.

PR #13, opened 2016-04-19
Add connections between the user, game, and piece models

obulkin:
- Added the Database Cleaner gem and configured RSpec to clean out the test DB after each run by calling it
- Added several associations and an active scope to the game model
- Added a method to the user model for listing a specific user's games
- Added an association to the piece model
- Added a method to the piece model for finding the owner of a piece
- Added foreign key constraints to the games table
- Added a foreign key constraint with cascading deletion to the piece table

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,6 +62,9 @@ group :development, :test do
   gem "factory_girl_rails", "~> 4.0"
 end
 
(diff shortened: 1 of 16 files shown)


lasley:
Solid - even covered the tests for things commonly missed like ForeignKeys, scopes, associations. 👍

PR #14, opened 2016-04-20
Board

aryawinterfell:


app/views/static_pages/board.html.erb
```diff
@@ -0,0 +1,123 @@
+<!DOCTYPE html >
+<html>
+<style type="text/css">
```
lasley: Let's add this into a CSS assets file, likely for the board. It's best to keep everything in their little isolated niche so that it is easier to find things when it comes time for tweaks

Take a look at this article, which gets into the pipeline and all that - https://launchschool.com/blog/rails-asset-pipeline-best-practices

app/views/static_pages/board.html.erb
```diff
@@ -0,0 +1,123 @@
...
+ 	table.one {
+ 		border-collapse: collapse;
+ 	}
+ 	td.black {
+ 		background-color: #000608;
+ 	}
+ 	</style>
+	<body>
+	<div class="board">
+ 		<table class="one">
```
lasley: We've got a lot of repetition here. Maybe a lil somethin somethin with some loops? Take a look at Ruby Integer `.upto` and `.downto` methods

app/views/static_pages/board.html.erb
```diff
@@ -0,0 +1,123 @@
...
+ 		 		<td class="black"></td>
+ 		 		<td></td>
+ 		 		<td class="black"></td>
+ 		 		<td></td>
+ 		 		<td class="black"></td>
+ 			</tr>
+ 			</table>
+ 		</div>
+	</body>
+</html>
```
lasley: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

app/views/static_pages/board.html.erb
```diff
@@ -0,0 +1,123 @@
+<!DOCTYPE html >
+<html>
+<style type="text/css">
+ 	.board {
+ 		width: 640px;
```
lasley: Take a look at `em` http://kyleschaeffer.com/development/css-font-size-em-vs-px-vs-pt-vs/

obulkin:
I second all the comments from @lasley. In addition, I'm not seeing a controller that corresponds to the route you created. This is needed to render the view.

aryawinterfell:
Oh okay, I figured I didn't have to make a controller bc this was just a temp fix to see the code for the board .. but then why would I have made the route? Duh. Thank you for all the links and stuff Dave, I'll clean her up ;)

app/views/static_pages/board.html.erb
aryawinterfell: A little confused about this... where am I suppose to and a \n ?

lasley: Just add a carriage return after `</html>`. Basically just need a blank line at the end of the page ;)

app/views/boards/show.html.erb
```diff
@@ -0,0 +1,28 @@
+<body>
+	<div class="board">
+ 		<table class="one">
+ 			<% 4.times do %>
+ 			<tr>
+ 		 		<td class="black"></td>
```
lasley: Hahah you're gonna hate me- but check out CSS nth-child operator, which will help reduce repetition further https://davidwalsh.name/css-tables-css3-alternate-row-colors

Something like this on the rows

```
<tr class="even">
...
<tr class="odd">
```

Then in Css, something like

```
.board tr.even td:nth-child(even) { background-color: #000; }
.board tr.even td:nth-child(odd) { background-color: #FFF; }
##  Then the opposite for tr.odd
```

Orrrrrrr if you wanted to solve this in Ruby- you could use modulo, which will allow you to determine even/odd

```
>>> 5%2
1
>>> 6%2
0
```

General pointer- if you are ever copy/pasting, solve it with code simplification instead ;)

lasley: Hmmm now that I think about it - maybe code complication depending on the way you look at it lol

aryawinterfell: No no that's good! Will go back to the drawing board lol

app/controllers/boards_controller.rb
```diff
@@ -0,0 +1,6 @@
+class BoardsController < ApplicationController
+
+	def show
+	end
+
+end
```
obulkin, line 6: Ruby files are generally indented using two spaces rather than a tab. I realize this sounds nitpicky, but it's a very common convention, so you may catch some flak for not following it. You can set up Sublime to enforce this for you by opening up a Ruby file, going to your syntax specific settings, and adding: 

```
"tab_size": 2,
"translate_tabs_to_spaces": true 
```

With these settings in place, whenever you use a tab in an `.rb` file, it'll automatically get converted into two spaces.

obulkin, line 6: On a related note, spaces are generally preferred to tabs for indenting code because they're treated more consistently. For example, GitHub tends to render tabs as being 8 spaces wide, whereas Sublime seems to default to 4. This is why your HTML and CSS files look much more deeply indented here.

spec/helpers/boards_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+#
+# describe BoardsHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
+RSpec.describe BoardsHelper, type: :helper do
+  pending "add some examples to (or delete) #{__FILE__}"
```
obulkin: Could you remove this line? It'll generate a reminder about pending tests every RSpec run, which we don't really need since we won't actually be adding tests here.

obulkin:
Could you please add a `.DS_Store` line to your `.gitignore`? This is a file type that gets generated automatically by the Finder app on OS X and shouldn't be committed to the repo.

app/assets/stylesheets/boards.scss
```diff
@@ -0,0 +1,27 @@
+ 	.board {
+ 		width: 60%;
+    	height: 70%;
+    	border: 3px solid #333;
+    	margin: 10px auto 10px auto;
+
```
obulkin: I just tried visiting this view in dev, and the alignment between the `.board` div and the table inside it is very brittle because the former uses `%` to adjust to the size of the viewport while the latter consists of fixed-size cells. At larger viewport sizes the border on the div goes out way past the board, while at small sizes, the div constrains the board and distorts it into a rectangle. 

This can be addressed by turning the wrapping div into a basic Bootstrap container and applying all the relevant properties (border and margin) directly to the table.

app/assets/stylesheets/boards.scss
```diff
@@ -0,0 +1,27 @@
...
+ 	}
+ 	table.one {
+ 		border-collapse: collapse;
+ 	}
+ 	.board tr.odd td:nth-child(odd){
+ 		  background-color: #000608
+ 	}
+ 	.board tr.even td:nth-child(even){
+ 		background-color: #000608;
+ 	}
```
obulkin: The ordering of black and white squares that results from this CSS is actually 90 degrees off from what a standard chessboard looks like. Whether you're playing as black or white, the lower left-hand corner should always be black, not white: https://en.wikipedia.org/wiki/Chessboard

obulkin:
Keep at it! 😄 This is looking much better, but it still needs some tweaks here and there, as indicated by my inline comments.

app/controllers/boards_controller.rb
aryawinterfell, line 6: Oh okay, that's really good to know actually. I was wondering about that.

app/assets/stylesheets/boards.scss
aryawinterfell: In the ticket it said to not use bootstrap, I thought about it myself. I think now that I built it without bootstrap maybe we can check with Dave and see if I can use now ?

obulkin: I think he primarily wanted us to avoid using Bootstrap columns/rows for the actual board, but it should be fine to wrap all the content in a Bootstrap container. In any event, I would still apply the border and margins directly to the table rather than a wrapper div.

app/assets/stylesheets/boards.scss
```diff
@@ -0,0 +1,34 @@
...
+ 	padding-top: 10%; 
+ 	padding-bottom: 10%;
+}
+td {
+  empty-cells: show;
+}
+table.one, .board {
+  border-collapse: collapse;
+ 	height: 45em;
+ 	width: 45em;
```
obulkin: Right here, you're assigning a height and width of 45em to the `.board` div, but on lines 2 and 3, you're setting those properties equal to 95%. Only one set of properties will be applied (the one that comes last), so you shouldn't leave them both in the code.

This leads me to ask why you're still wrapping the table in a `board` div. It doesn't give you anything that can't be accomplished by styling the table directly and just causes problems, which you've managed to deal with right now, but which may crop up again when we make the board responsive. Ultimately, it's up to you, and I won't hold up the merge because of this choice, but I think it'll just cause unnecessary trouble.

app/assets/stylesheets/boards.scss
```diff
@@ -0,0 +1,34 @@
...
+}
+.board tr.even td:nth-child(even){
+ 	background-color: #000608;
+}
+.container{
+ 	width:auto;
+  height:auto;
+  max-width:90%;
+  display: table;
+}
```
obulkin, line 18: `.container` is one of the core classes provided by Bootstrap and will ultimately be used all over the site. You shouldn't be modifying its properties directly because other people will use it elsewhere with the expectation that it will conform to the Bootstrap standards.

app/assets/stylesheets/boards.scss
```diff
@@ -0,0 +1,34 @@
...
+body {
+ 	padding-top: 10%; 
+ 	padding-bottom: 10%;
+}
+td {
+  empty-cells: show;
+}
+table.one, .board {
+  border-collapse: collapse;
+ 	height: 45em;
```
obulkin: Optional: I'm seeing a weird issue where some of the squares are slightly smaller than others. I think this may be due to rounding errors stemming from the fact that you're providing dimensions for the board as a whole and not for each individual cell. Doing the latter and just allowing the table to adjust accordingly seems to take care of it.

app/views/boards/show.html.erb
```diff
@@ -0,0 +1,28 @@
...
+         <% 8.times do %>
+ 	        <td></td>
+         <% end %>
+ 			  </tr>
+ 			
+      <% end %>
+
+	  </table>
+  </div>
+ </div>
```
obulkin: It looks like you're still having some issues with inconsistent indentation. Could you try adding `"translate_tabs_to_spaces": true` to your user-level Sublime settings? This should allow you to use the tab key everywhere (much easier than hitting space a bunch of times), while producing files that are always indented in terms of spaces.

obulkin:
Almost there! Please see my comments in line. Also, now that you've added `.DS_Store` to the .gitignore, could you please remove that file from the branch?

app/assets/stylesheets/boards.scss
aryawinterfell: Don't I need to have a border around the board? That's what the board div is doing.

app/views/boards/show.html.erb
aryawinterfell: Ya, it's really annoying that it happened again. Much easier now with they syntax override, but obviously it's still off.

aryawinterfell:
Which file exactly did you want me to delete?

On Thu, Apr 28, 2016 at 9:58 PM, Oleg Bulkin notifications@github.com
wrote:

> Almost there! Please see my comments in line. Also, now that you've added
> .DS_Store to the .gitignore, could you please remove that file from the
> branch?
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> https://github.com/fp-team-maniac/maniac-chess/pull/14#issuecomment-215610704

## 

www.molello.com

347.861.8591

app/assets/stylesheets/boards.scss
obulkin: Any HTML element can have a border, so you can just put one directly on the table.

obulkin:
You should remove `.DS_Store`. It's just a file created automatically by Finder and doesn't need to be in the repo.

lasley:
(FYI `.` prefixed files are hidden on OSX - `CMD`+`Shift`+`.` will show them)

app/assets/stylesheets/boards.scss
```diff
@@ -0,0 +1,19 @@
+body {
+padding: 10%;
+width: 60em;
+margin: 0 auto 0 auto;
```
obulkin: The `<body>` element effectively represents the entire visible page that all the other elements live on, so it generally shouldn't have margins or a specific width. Padding may make sense in some cases but isn't necessary here since we're using Bootstrap containers, which come with their own padding. Moreover, it would interfere with a top nav, which we'll probably want to add eventually.

Aside from these points, styling the `<body>` in this way should also be avoided for the reason I mentioned earlier in the context of the `.container` class: these CSS rules will apply across the entire site and every page on the site will have a `<body>`.

aryawinterfell: I think what you're trying to get at is that you want me to just use bootstrap to center the board. Right?

obulkin:
@aryawinterfell --- I wasn't suggesting that you use Bootstrap to center the board. There may be a Bootstrap class that you can use to accomplish this, but the easiest way to center an element horizontally relative to its parent element is just to give it left and right margins of `auto`. In this particular case, the element you want to center is `<table>`, and the parent is the `.container` div, which is itself centered relative to `<body>` thanks to Bootstrap.

Your current approach doesn't actually work because the board will be left-centered inside the Bootstrap column. Also, you're using the `style` attribute to add CSS properties directly to the HTML, which is generally considered a bad practice. A single element can have many classes on it, so if you want to add some extra styling to an element with Bootstrap classes, one good way to do that is to add your own class to that element as well and tie your extra style rules to that custom class.

lasley:
https://jsfiddle.net/yoj22548/

obulkin:
Thanks for the JSFiddle, @lasley. That's exactly what I had in mind.

app/views/boards/show.html.erb
```diff
@@ -0,0 +1,21 @@
+<body>
+ <div class="container">
+ <table class="block-center"
```
obulkin: You forgot to end this tag with `>`.

app/views/boards/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+   <table class="one">
```
obulkin: You can nest tables by putting one table inside a cell (the `<td></td>` portion) belonging to another table, but the current approach will not work and will actually result in the two getting rendered one after the other. The reason you're not seeing this is because you didn't close the table tag on the preceding line and the browser is treating them as one element.

The solution is just to drop one of the `<table></table>` pairs. You only need one to make this work.

app/views/boards/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <tr class="odd">
+       <% 8.times do %>
+ 	       <td></td>
+       <% end %>
+    	</tr>
+ 	   <% end %>
+    </table>
+   </table>
+  </div>
+ </div>
```
obulkin: There's only one opening div tag at the top but you have two closing div tags at the bottom.

config/routes.rb
```diff
@@ -1,5 +1,8 @@
 MANIACChess::Application.routes.draw do
   devise_for :users
+ resource :board, only: [:show]
```
obulkin: Informational: When you're setting up a resourceful route for a single action, you don't need use the array notation. In other words, this line can just be `resource :board, only: :show`.

obulkin:
Please see my comments in line. You're definitely on the right path now, but the HTML is messy and needs to be cleaned up a bit.

Also, you'll need to get the branch ready for a merge by merging in master and resolving any conflicts.

app/views/boards/show.html.erb
aryawinterfell: Was running out the door when I did this and didn't check the html, should have done that!! Sorry for silly syntax mistakes.

obulkin:
Awesome! This looks good to go now. 👍  You just need to resolve the merge conflicts, so that I can merge it in. If you need a refresher on how to do that, check out this FP video: https://www.thefirehoseproject.com/advanced-git-topics/5

aryawinterfell:
How do you handle conflicts in the routes? Do I just add the route for the board ?

lasley:
I'll hit you up in Slack and we can go over it

lasley:
👍

PR #15, opened 2016-04-22
CRUD for games model and controller set up and tested

AidanZeber:
I've set up the basic CRUD functionality within the "games" controller, set up simple form within the new game page, and created some basic dummy views (index, show, new). All tests passing on my machine. Please review.


app/controllers/games_controller.rb
```diff
@@ -0,0 +1,34 @@
+class GamesController < ApplicationController
+
+  def new
+    @game = Game.new
+  end
+
+  def index
```
obulkin: Based on the wireframes, it doesn't actually seem like we need an index action, so I would remove it and the associated view.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def create
+    @game = Game.create(game_params)
+    @game.update_attribute(white_player_id, current_user.id)
```
obulkin: There are a number of problems here:
- Expecting `create` to fail and using it anyway is not a good idea because it will connect to the DB and create a transaction that will then need to be rolled back. If you want to build a model object without saving it, you should just use `new`.
- `update_attribute` is a dangerous method because it bypasses validations (http://stackoverflow.com/questions/2778522/rails-update-attribute-vs-update-attributes). Unless you're totally sure you want that behavior, just use `update_attributes` instead.
- There are validations on the game model that could cause record creation to fail, so you should check for this scenario and deal with a validation failure appropriately.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    redirect_to root_path
+  end
+
+  def show 
+    @game = Game.find_by_id(params[:id])
```
obulkin: Although it's tempting, I would avoid using variants of `find_by` like `find_by_id` or `find_by_name`. They aren't defined explicitly and instead take advantage of Ruby's process for looking up missing constants, which makes them inefficient. If you're looking things up just based on ID, `find` is what you need, and if you're using a different property, then the syntax is `find_by(property: value)`.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    if @game.blank?
+      render text: 'Not Found', status: :not_found
+    end
```
obulkin: This is just a style point, but I generally try to use the single-line, trailing `if` in scenarios like this: `do_stuff if time_is_right?`.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def update 
+    @game = Game.find(params[:id])
+    @game.update_attributes(black_player_id: current_user.id)
+    redirect_to game_path(@game.id)
+  end
```
obulkin, line 27: When users create a game, we should probably allow them to choose whether they want to play as black or white, which means that here the code should figure out which seat is empty and assign the second user to that color.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  private
+
+  def game_params
+    params.require(:game).permit(:name, :black_player_id)
```
obulkin: Since player IDs will always be coming from Devise via `current_user`, there's no need to whitelist a player ID key.

config/routes.rb
```diff
@@ -1,5 +1,8 @@
 MANIACChess::Application.routes.draw do
   devise_for :users
+
+  resources :games
```
obulkin: You're creating a number of routes here that aren't supported by the controller. Could you narrow this down using `only` or `except`?

config/initializers/simple_form.rb
```diff
@@ -1,5 +1,6 @@
 # Use this setup block to configure all options available in SimpleForm.
 SimpleForm.setup do |config|
+  config.input_class = "form-control"
```
obulkin: This option should be redundant because it's part of the `:vertical_form` wrapper defined in simple_form_bootstrap.rb, which should in turn be the default wrapper. Did you find that adding it actually changes the resulting HTML?

spec/controllers/games_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  describe "games#new action" do 
+    it "should sucessfully show the new form" do
+      get :new
+      expect(response).to have_http_status(:success)
+    end
+  end
+
+  describe "games#show action" do 
+    it "should successfully show the page if the game is found" do
+      game = FactoryGirl.create(:game)
```
obulkin: FYI, I've configured RSpec in such a way that you can simply use `create(:game)` here without the module name. But this is totally fine too. 😄

spec/controllers/games_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  describe "games#index action" do
+    it "should successfully show the page" do 
+      get :index
+      expect(response).to have_http_status(:success)
+    end
+  end
+
+  describe "games#new action" do 
+    it "should sucessfully show the new form" do
```
obulkin: Your test names are a bit confusing because the tests don't actually check whether a particular form or page is being shown. Instead, they're just looking for the server response to include a certain status. Could you rename them accordingly?

app/controllers/games_controller.rb
AidanZeber: What is the disadvantage of leaving it? Understood that we may not have a need for it right now according to our wireframes, however, this is a basic aspect of many applications and it may be useful in the future.

obulkin: These points don't apply strongly in this particular case because this is a few lines of code in a student app, but generally speaking, merging code that won't be needed in the foreseeable future seems like a bad practice for the following reasons:
- It adds to the complexity of the app and the burden associated with maintaining it. The excess code can break, causing exceptions, and it has associated tests, so even if the breakdown is more subtle, it could hold up a deployment or a merge.
- If the code is user-facing, which this is, it could easily result in bad/confusing UX.

app/controllers/games_controller.rb
lasley: While generally I would agree, the multi-line is more clear here IMO. My general rule of thumb is to not compound multiple expressions on the same line. 

```
render text: 'Not Found', status: :not_found if @game.blank?
```

In the above, it is not clear from a glance where the method ends and the condition begins. 

In other languages this would probably be fine, but due to the Ruby style point of not using `(` and `)` for most methods, we loose the clarity. 

The below would be more clear, but it would break the no parenthesis styling

```
render (text: 'Not Found', status: :not_found) if @game.blank?
```

lasley: On second glance, the first one is being applied like so:

```
render (text: 'Not Found', status: (:not_found if @game.blank?) )
```

app/controllers/games_controller.rb
lasley, line 27: We wouldn't want to do any sort of logic like that in the controller, that would be where the model comes in. 

Choice of color isn't an explicit requirement of the game though, so maybe a ticket & subsequent team vote to add it? Or has this already been decided via wireframes/other convos, and I'm just forgetting things?

app/views/games/index.html.erb
```diff
@@ -0,0 +1 @@
+<h1> Games </h1>
```
lasley: All files should end with a carriage return for POSIX compliance (one blank line at end of file)

http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

spec/controllers/games_controller_spec.rb
lasley: `it "should return success response on get request"`

app/controllers/games_controller.rb
AidanZeber: That makes sense. Will remove.

app/controllers/games_controller.rb
obulkin: Fair enough. There's plenty of room here for personal/team preference. I find the single line, trailing conditionals pretty elegant compared to the standard version and tend to use them a lot, but I can totally see the concerns around clarity, especially when methods without parens are also involved.

app/controllers/games_controller.rb
obulkin, line 27: Good points. I don't think this has really been decided yet. To keep things simple, we can just assume for now that the first player will always be seated as white and the second player will play as black. I'll add a ticket to the icebox for allowing users to select a color when they create a game, and we can work on it if/when we see fit.

app/controllers/games_controller.rb
AidanZeber: So we can remove the "black_player_id" reference in both the update and game_params methods?

obulkin: Well to some extent it depends on the use case that `update` is meant to handle. My understanding from reading through the code is that you intended to have it support the scenario where a second player is joining a game with an open seat. If that's the case, and we go with the simple approach of always assigning the first user to be the white player, then `black_player_id` still needs to be updated as part of the action.

However, it doesn't need to be part of `game_params`. The point of `params.require` and `params.permit` is to whitelist certain values that come through the params hash for use in record creation/modification. For security reasons, they can't be used this way by default: http://guides.rubyonrails.org/action_controller_overview.html#strong-parameters. Since the player ID is coming from Devise here rather than from the params hash, there's no need for it to be whitelisted.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,38 @@
+class GamesController < ApplicationController
+  before_action :authenticate_user!, only: [:new, :create, :edit, :update, :destroy]
+
+  def new
+    @game = Game.new
+  end
+
+  def create
+    @game = Game.create!(game_params.merge(white_player_id: current_user.id))
```
obulkin: As with `save!`, `create!` will throw an exception if there's a validation error (http://guides.rubyonrails.org/active_record_basics.html#validations), and this method is not set up to rescue from it, so you probably want to use `create` instead.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    if @game.save
```
obulkin: `create` and its bang counterpart both generate an object and attempt to save it, so calling `save` here is redundant. Instead, it should suffice to check whether the object was actually saved as part of the `create` call by calling `persisted?`.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,38 @@
+class GamesController < ApplicationController
+  before_action :authenticate_user!, only: [:new, :create, :edit, :update, :destroy]
```
obulkin: Optional: There are no `edit` or `destroy` actions in the controller, so they don't need to be mentioned on this line.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    else
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  def show
+    begin 
+      @game = Game.find(params[:id])
+    rescue
+      if @game.nil?
```
obulkin: Generally speaking, you should not use `rescue` without providing a specific exception class. Rails and Ruby both use exceptions to communicate a wide range of failures, so a blanket `rescue` could result in strange behavior and lots of debugging issues by masking all sorts of problems.

The particular exception class you're looking for here is `ActiveRecord::RecordNotFound` (http://guides.rubyonrails.org/active_record_querying.html#retrieving-a-single-object) and the syntax is just `rescue ActiveRecord::RecordNotFound`. This approach also has the added benefit of making the conditional on the next line unnecessary since `render` will now be called only when the game cannot be found.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+        render text: 'Game Not Found', status: :not_found
+      end
+    end
+  end
+
+  def update 
+    @game = Game.find(params[:id])
+    @game.update_attributes(current_user.id)
```
obulkin: This is the wrong syntax for `update_attributes`, which requires key-value pairs, just like `create` and `new`.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    redirect_to game_path(@game)
+  end
+
+  private
+
+  def game_params
+    params.require([:state]).permit(:name)
```
obulkin: This should say `:game` instead of `:state`. I believe it used to be correct, so I'm guessing this is a left-over from all of the test debugging that happened recently.

config/routes.rb
```diff
@@ -1,5 +1,8 @@
 MANIACChess::Application.routes.draw do
   devise_for :users
+
+  resources :games, :only => [:new, :show, :create]
```
obulkin: You're missing an `update` route here.

spec/factories/games.rb
```diff
@@ -5,4 +5,8 @@
     association :black_player, factory: :user
     state "open"
   end
+  
+  factory :invalid_game, parent: :game do
```
obulkin, line 9: FYI: If you nest the child factory inside the do/end block used to create the parent, then you don't have to specify the parent explicitly using the `parent:` option.

obulkin:
This is looking better, but I'm still seeing some issues here and there. Please see my inline comments for more info.

Also, I don't see any tests for the `update` action. Were these intentionally left out as well?

app/controllers/games_controller.rb
AidanZeber: I've tried both forms of the create function (with and without !), errors result either way and end up in failure. I've also tried using a version of .new, still no luck.

app/controllers/games_controller.rb
AidanZeber: Removed to only reflect actual actions within the controller.

app/controllers/games_controller.rb
AidanZeber: I've tried it both ways, at the suggestion of my mentor we changed from :game to :state. Both still ultimately fail the create tests.

app/controllers/games_controller.rb
AidanZeber: Got it, added "black_player_id:" to the call, the update function should have the second player be the black player, create function assigns to white.

app/controllers/games_controller.rb
obulkin: What sort of errors are you getting when you use `create`? If it's called properly and with valid data, it should not generate any exceptions, whereas `create!` will definitely produce an exception when validations fail. Since it's normal for validations to fail in production, this is a big distinction, and you shouldn't use `create!` unless the method is set up to rescue the resulting exception.

In other words, just go with `create`. 😄

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    else
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  def show
+    begin 
+      @game = Game.find(params[:id])
+    rescue ActiveRecord::RecordNotFound
+      if @game.nil?
```
obulkin: As I mentioned in my earlier comment, there's no need to check if `@game == nil`. Since we're now inside the `rescue` block, you can be confident that an `ActiveRecord::RecordNotFound` exception was thrown by `find` and that it's appropriate to generate a 404 using `render`.

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    @game = Game.create(game_params.merge(white_player_id: current_user.id))
+    if @game.persisted?
+      redirect_to game_path(@game)
+    else
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  def show
+    begin
```
obulkin: Informational: Ruby methods implicitly act like a `begin` block, so you don't need to use `begin` to rescue from errors thrown anywhere in the method: http://seejohncode.com/2012/04/17/short-tip-rescuing-a-method-in-ruby/

spec/controllers/games_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+RSpec.describe GamesController, type: :controller do
+
+  describe "games#new action" do 
+    it "should return success response on the get request" do
+      user = create(:user)
+      sign_in user
+      get :new
+      expect(response).to have_http_status(:success)
+    end
+  end
```
obulkin, line 17: Could you add a test for the `Games#new` behavior when a user is not signed in?

obulkin:
Why are your tests for `Games#update` commented out? Are you unable to get them working as well?

spec/controllers/games_controller_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    it "should return a Game Not Found error if the game is not found" do 
+      get :show, id: '99999999'
+      expect(response).to have_http_status(:not_found)
+    end
+  end
+
+  describe "games#update action" do 
+    before :each do 
+      @game = FactoryGirl.create(:game, name: "Test Name", state: "open")
+    end
```
obulkin: When possible, it's generally considered better to use `let` in these scenarios rather than `before` and an instance variable: http://stackoverflow.com/questions/5359558/when-to-use-rspec-let. See here for info on `let`: http://www.relishapp.com/rspec/rspec-core/v/3-4/docs/helper-methods/let-and-let

app/controllers/games_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def show 
+    @game = Game.find(params[:id])
+    rescue ActiveRecord::RecordNotFound
+    render text: 'Game Not Found', status: :not_found
+  end
+
+  def update 
+    @game = Game.find(params[:id])
```
obulkin: Optional: A general principle that I like to adhere to is making data and methods only available to the parts of the app that need them. In this case, since you're not trying to relay data to a view (the normal use for instance variables in controllers), you can make do with just a local version: `game`.

spec/controllers/games_controller_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+    context "with valid attributes" do
+      it "located the requested @game" do
+        user = create(:user)
+        sign_in user
+        put :update, id: @game, game: FactoryGirl.attributes_for(:game)
+        expect(@game).to eq(@game)
```
obulkin: Right here, you're just comparing the `@game` variable created as part of the spec to itself, which is not very useful. 

My approach would be just to roll this into the next test. If the action is properly updating a specific game's attributes, then it must be retrieving the correct game in the first place.

spec/controllers/games_controller_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    end
+
+    it "does not change @game's attributes" do 
+      user = create(:user)
+      sign_in user
+      put :update, id: @game, 
+        game: FactoryGirl.attributes_for(:game, name: "Test Game", state: nil)
+        @game.reload
+        expect(@game.name).not_to eq("Test Game")
+        expect(@game.state).to eq("open")
```
obulkin: Your tests for the `update` action don't align with what it's actually doing. The only property of the game that it's changing is `black_player_id`, and it's doing this based on info from Devise, not data that's coming through the `params` hash. 

The general pattern should be as follows:
- Create and sign in a user
- Update the game, which should be created with two different users
- Make sure that the game's black player is now the user that you signed in

spec/controllers/games_controller_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  describe "games#create action" do
+    context "with valid attributes" do
+      it "creates a new game" do 
+        user = create(:user)
+        sign_in user
+        expect{
+          post :create, game: attributes_for(:game)
+        }.to change(Game,:count).by(1)
```
obulkin: As part of this test, could you also confirm that the new game has the attributes that you passed in? Using attributes that are not part of the standard factory might help out with doing that in a reliable way.

spec/controllers/games_controller_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      end
+
+      it "redirects to the new game" do
+        user = create(:user)
+        sign_in user
+        post :create, game: FactoryGirl.attributes_for(:game)
+        expect(response).to redirect_to Game.last
```
obulkin: Optional: I'm not a big fan of using `first` and `last` in tests, because that makes the results dependent on the DB being in a clean state at the beginning of the test. I would go with a distinct name during the creation step and then a `find_by` that references the name.

Also, this should probably use the path helper for the `show` action. Your current syntax may match to a variety of paths. The API is kind of unclear on this point: http://api.rubyonrails.org/classes/ActionDispatch/Assertions/ResponseAssertions.html.

obulkin:
Keep chugging along! The core code looks good now, but the tests still need some work, especially the ones for `Games#update`, which currently don't really align with what the action does.

spec/controllers/games_controller_spec.rb
AidanZeber: This is what I believe is a fairly standard approach to testing PUT methods. I understand limiting the number of tests and rolling the this into the next test, but I as far as locating the correct game before making any updates this seems correct to me.

http://everydayrails.com/2012/04/07/testing-series-rspec-controllers.html

obulkin: Well really, the main problem is that you're comparing the instance variable `@game` created in the `before` block to itself. In the article that you shared, the writer is using the `assigns` hash (http://guides.rubyonrails.org/testing.html#the-four-hashes-of-the-apocalypse) to compare the instance variable created in the spec to the one created in the controller action. That would work here but isn't compatible with my suggestion to just use a local variable in the action, which is why it seems easiest to simply roll this into the next example.

Another reason for doing so is that testing a specific controller variable would couple the test too closely to the controller implementation. Ultimately, it doesn't matter what intermediate variables are used to accomplish the task as long as the action properly updates the correct game, so the tests should aim to establish this in the least brittle way possible.

obulkin:
@AidanZeber --- I just reviewed this again and there are still two inline comments around the `Games#create` unit tests that haven't been addressed. Should I make those changes and push to the branch? I don't think it makes sense to move small tweaks like this into a separate Trello ticket.

lasley:
@obulkin - If you can fix them before merge, please do.

obulkin:
Will do.

obulkin:
Since the unit tests for the `update` action have been moved into a separate ticket, and the other issues have been addressed, this is now good to go. I'll go ahead and merge the branch.

PR #16, opened 2016-04-26
Piece model method for identifying obstructed moves

obulkin:


app/models/piece.rb
```diff
@@ -10,4 +10,78 @@ class Piece < ActiveRecord::Base
...
+  # This method only checks for obstructions along horizontal, vertical, and diagonal moves. Before calling it, 
+  # please make sure that the move provided falls into one of these categories and that an obstruction check is 
+  # actually needed (e.g. the piece isn't a knight, the move is not malformed).
+  def is_move_obstructed?(new_rank, new_file)
+    current_pieces = game.pieces
+
+    # These are assumed to be diagonal moves along a slope of 1
+    if new_rank > rank && new_file > file
+      (1..new_rank - rank).each do |offset|
+        piece_at_offset = current_pieces.select{|piece| piece.rank == rank + offset && piece.file == file + offset}
```
lasley, line 28: Might make sense to have a `get_piece_at(rank, file)` method to reduce some of this code repetition.

app/models/piece.rb
```diff
@@ -10,4 +10,78 @@ class Piece < ActiveRecord::Base
...
+        # Exception for an enemy piece in the spot at the end of the move
+        next if offset == new_rank - rank && piece_at_offset.present? && piece_at_offset.first.color != color 
+        return true if piece_at_offset.present?
+      end
+      false
+    elsif new_rank < rank && new_file < file
+      (new_rank - rank..-1).each do |offset|
+        piece_at_offset = current_pieces.select{|piece| piece.rank == rank + offset && piece.file == file + offset}
+        next if offset == new_rank - rank && piece_at_offset.present? && piece_at_offset.first.color != color
```
lasley: Maybe something like below to reduce repetition

```
skip_enemy = ->(offset, new_offset, piece) { 
  offset == new_offset && piece.present? && piece.first.color != color 
}
```

lasley:
The logic is sound, and tests back it up. Few stylistic choices, but no blockers 👍

obulkin:
@lasley --- Thanks for the review. Your suggestions both amount to abstracting a single recurring line into a separate method, which doesn't seem like a significant improvement to me (instead of 8 nearly identical lines, there will now be 8 nearly identical method calls, plus an extra helper).

I could also try to abstract the entire iteration block into a helper like

```
def evaluate_square(current_pieces, eval_rank, eval_file, offset, peak_offset)
  piece_at_offset = current_pieces.select{|piece| piece.rank == eval_rank && piece.file == eval_file}
  return false if offset == peak_offset && piece_at_offset.present? && piece_at_offset.first.color != color
  return true if piece_at_offset.present?
  false
end
```

and then do

```
return true if evaluate_square(current_pieces, rank + offset, file + offset, offset, new_rank - rank)
```

That would start to dry things up significantly, but I'm hesitant to do it because the large number of parameters makes the code harder to follow.

lasley:
Agreed too many params on the one method, although you could probably get rid of `current_pieces` in favor of a `game.pieces`.

Reasoning behind the two methods (and the use of a lambda for the `skip_enemy` instead of a true method) is that the `get_piece_at` method could be useful outside of this context, such as when a piece is being captured. The offset skipper isn't useful outside of the context of collision detection IMO

obulkin:
@lasley --- Hmm. I can see the potential reuse associated with `get_piece_at`, but since it abstracts just a single, fairly straightforward expression, I would imagine that a different person working in the codebase isn't likely to stop and think "Has this problem been solved already?" when they run into a relevant scenario. That means that the reuse would need to be enforced.

How are these considerations normally addressed in a professional environment?

obulkin:
Also, if I go with a lambda literal for skipping enemies, wouldn't I have to repeat the declaration in each loop, making the end result basically the same?

lasley:
Excellent point @obulkin - we can't fully enforce it, because like you said it is simply an expression. This is where peer reviews come in though - typically someone is going to see the logic and point towards a refactor. Sometimes it goes unnoticed, merges in, and someone consolidates later (this is what happens most of the time in open source).

Let's say we determine there is a bug in the way the query is being run (probably not the case here, but over-generality for the sake of conversation). How many spots do we now need to change the expression in, and how many tests now need to be updated?

As I'm thinking about it, it might make sense to just all the method `at_pos` as adding piece would be redundant naming in the piece model.

Regarding the lambda, you can just declare it at the top of the method and use it throughout. Can just call the variable like a method, providing the params.

lasley:
An example of a very common helper similar to this in a lot of projects, that nearly everyone forgets to use `Person.get_by_name` vs a lookup and querying the name

lasley:
Although now that I'm thinking about it, I think ActiveRecord creates those simplistic helpers for you automatically so my point is kind of invalid there. More valid elsewhere without the magic, but yeah there's plenty of stupid things like that

obulkin:
@lasley --- Another concern associated with `get_piece_at` is that I'm intentionally using `select` here to prevent `is_move_obstructed?` from hitting the DB a dozen times, but there may be other situations where forcing a fresh result with `where` is preferable, and making this a project-wide helper would mask that decision.

I do like the lambda idea, though. That line is kind of unclear right now (hence my comment), and a lambda with a good variable name would make it easier to understand.

lasley:
@obulkin Good point with the global helper, which would nullify my capture example because you would want to guarantee a non-stale result. Doesn't make sense to try and refactor for unknown abstraction, vs just moving something to a different scope.

I think the lambda may cut the real issue of legibility anyways, I just had to stare at those section for a bit. I'd say gtg with just the lambda; maybe breaking the query into one too but that may be excessive as it's a lot more simple than the skip expression.

obulkin:
Yeah. I considered putting the select expression into a lambda as well, but I think that line is already fairly clear thanks to the `piece_at_offset` variable.

lasley:
👍

AidanZeber:
Reviewed code and all inline comments. Branch approved. Merging branch.

PR #17, opened 2016-05-05
added king valid move

brandoncha:
Please see and let me know if it needs to modify.

### DIFF
--- a/app/models/king.rb
+++ b/app/models/king.rb
@@ -1,2 +1,11 @@
 class King < Piece
+	def valid_move?(new_rank, new_file)
+		rank_diff = (new_rank - rank).abs
+		file_diff = (new_file - file).abs
+		if (rank_diff <= 1) && (file_diff <= 1) && !is_move_obstructed?(new_rank, new_file) && !is_move_malformed?(new_rank, new_file)
+			return true			
+		else
+			return false
+		end	
+	end
 end
--- a/app/models/king.rb
+++ b/app/models/king.rb
@@ -2,10 +2,10 @@ class King < Piece
 	def valid_move?(new_rank, new_file)
 		rank_diff = (new_rank - rank).abs
 		file_diff = (new_file - file).abs
-		if (rank_diff <= 1) && (file_diff <= 1) && !is_move_obstructed?(new_rank, new_file) && !is_move_malformed?(new_rank, new_file)
-			return true			
+		if !is_move_malformed?(new_rank, new_file) && (rank_diff <= 1) && (file_diff <= 1) && !is_move_obstructed?(new_rank, new_file)
+			true			
 		else
-			return false
+			false
 		end	
 	end
 end
--- a/app/models/king.rb
+++ b/app/models/king.rb
@@ -1,11 +1,11 @@
 class King < Piece
-	def valid_move?(new_rank, new_file)
-		rank_diff = (new_rank - rank).abs
-		file_diff = (new_file - file).abs
-		if !is_move_malformed?(new_rank, new_file) && (rank_diff <= 1) && (file_diff <= 1) && !is_move_obstructed?(new_rank, new_file)
-			true			
-		else
-			false
-		end	
-	end
+  def valid_move?(new_rank, new_file)
+    rank_diff = (new_rank - rank).abs
+    file_diff = (new_file - file).abs
+    if !is_move_malformed?(new_rank, new_file) && (rank_diff <= 1) && (file_diff <= 1) && !is_move_obstructed?(new_rank, new_file)
+      true      
+    else
+      false
+    end 
+  end
 end
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -90,7 +90,11 @@ class Piece < ActiveRecord::Base
   end
   
   def is_move_malformed?(new_rank, new_file)
-      rank == new_rank && file == new_file
+      return true if rank == new_rank && file == new_file
+      return true if rank < 1 || rank > 8
+      return true if file < 1 || file > 8
+      return true if new_rank < 1 || new_rank > 8
+      return true if new_file < 1 || new_file > 8
   end
   
 end
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -11,6 +11,10 @@ class Piece < ActiveRecord::Base
     color == "white" ? game.white_player : game.black_player
   end
 
+  def is_move_malformed?(new_rank, new_file)
+    rank == new_rank && file == new_file
+  end
+
   protected
   # This method only checks for obstructions along horizontal, vertical, and diagonal moves. Before calling it, 
   # please make sure that the move provided falls into one of these categories and that an obstruction check is 
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -11,10 +11,6 @@ class Piece < ActiveRecord::Base
     color == "white" ? game.white_player : game.black_player
   end
 
-  def is_move_malformed?(new_rank, new_file)
-    rank == new_rank && file == new_file
-  end
-
   protected
   # This method only checks for obstructions along horizontal, vertical, and diagonal moves. Before calling it, 
   # please make sure that the move provided falls into one of these categories and that an obstruction check is 
@@ -92,4 +88,9 @@ class Piece < ActiveRecord::Base
       false
     end
   end
+  
+  def is_move_malformed?(new_rank, new_file)
+      rank == new_rank && file == new_file
+  end
+  
 end
--- a/spec/models/king_spec.rb
+++ b/spec/models/king_spec.rb
@@ -4,18 +4,18 @@ RSpec.describe King, type: :model do
   let(:game) {create :game}
   
   describe "King's valid_move" do
-  	it "should properly evaluate blocked moves that are otherwise valid" do
+    it "should properly evaluate blocked moves that are otherwise valid" do
       king = game.pieces.find_by(rank: 1, file: 5, color: "white")
       king2 = game.pieces.find_by(rank: 8, file: 5, color: "black")
-  		expect(king.valid_move?(1, 6)).to eq(false)
-  		expect(king.valid_move?(1, 4)).to eq(false)
-  		expect(king.valid_move?(2, 5)).to eq(false)
+      expect(king.valid_move?(1, 6)).to eq(false)
+      expect(king.valid_move?(1, 4)).to eq(false)
+      expect(king.valid_move?(2, 5)).to eq(false)
       expect(king.valid_move?(2, 4)).to eq(false)
       expect(king.valid_move?(2, 6)).to eq(false)
       expect(king2.valid_move?(7, 5)).to eq(false)
       expect(king2.valid_move?(7, 4)).to eq(false)
       expect(king2.valid_move?(7, 6)).to eq(false)
-  	end
+    end
 
     it "shouldn properly evaluate moves that are not allowed for a king but are otherwise valid" do
       king1 = game.pieces.create(type: "King", rank: 3, file: 6, color: "white")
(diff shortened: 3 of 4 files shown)


app/models/king.rb
```diff
@@ -1,2 +1,11 @@
 class King < Piece
+	def valid_move?(new_rank, new_file)
+		rank_diff = (new_rank - rank).abs
+		file_diff = (new_file - file).abs
+		if (rank_diff <= 1) && (file_diff <= 1) && !is_move_obstructed?(new_rank, new_file) && !is_move_malformed?(new_rank, new_file)
```
obulkin: `is_move_obstructed?` is not designed to handle malformed moves and will generally be the most expensive procedure that needs to be called as part of `valid_move?`, so I would use the following order instead:
- Call `is_move_malformed?`
- Perform any type-specific validations
- Call `is_move_obstructed?` if necessary (it won't be for the Knight model)

app/models/king.rb
```diff
@@ -1,2 +1,11 @@
...
+			return true			
+		else
+			return false
```
obulkin: Optional: Ruby methods always return the value of the last statement in the method. In this case, the last line will be either `true` or `false` depending on the outcome of the conditional, which is what you want to return anyway, so it's not necessary to use an explicit `return`.

app/models/piece.rb
```diff
@@ -11,6 +11,10 @@ def user
     color == "white" ? game.white_player : game.black_player
   end
 
+  def is_move_malformed?(new_rank, new_file)
+    rank == new_rank && file == new_file
+  end
```
obulkin: It probably makes sense to have this method be protected as well, so I would move it down below the `protected` keyword. See here for more info on `public`, `private`, and `protected` in Ruby: http://stackoverflow.com/questions/3534449/why-does-ruby-have-both-private-and-protected-methods

spec/models/king_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  		game = create :game
+  		king2 = game.pieces.create(type: "King", rank: 4,  file: 5, color: "black")
+  		expect(king2.valid_move?(4, 6)).to eq(true)
+  		expect(king2.valid_move?(4, 4)).to eq(true)
+  		expect(king2.valid_move?(5, 6)).to eq(true)
+  	end
+  
+    it "shouldn't be on same sqaure after moving" do
+      expect(king.valid_move?(1, 5)).to eq(false)
+    end
```
obulkin: This will be taken care of by `is_move_malformed?`, which will have its own tests, so it's redundant to test it here.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  	it "should be inside destination, not blocked" do
+  		game = create :game
+  		king2 = game.pieces.create(type: "King", rank: 4,  file: 5, color: "black")
+  		expect(king2.valid_move?(4, 6)).to eq(true)
+  		expect(king2.valid_move?(4, 4)).to eq(true)
+  		expect(king2.valid_move?(5, 6)).to eq(true)
+  	end
+  
+    it "shouldn't be on same sqaure after moving" do
```
obulkin: This will be taken care of by `is_move_malformed?`, which will have its own tests, so it's redundant to test it here.

app/models/king.rb
brandoncha: I don't understand that `is_move_obstructed?` handles malformed moves here. It only covers obstruction moves. `is_move_malformed?` handles malformed moves here.

And do you suggests to do like this?

def valid_move?(new_rank, new_file)
    return false if is_move_malformed?
    return false if is_move_obstructed?
    (new_rank - rank).abs <= 1 && (new_file - file).abs <=1
end

app/models/piece.rb
brandoncha: I will move it down below the `protected` and let Ken know if he won't put this method under `protected'. thanks for the resource.

spec/models/king_spec.rb
brandoncha: yes right. i just tested it for my own version. I will delete it.

app/models/king.rb
obulkin: What I mean is that `is_move_obstructed?` may do something unexpected, like throwing an exception or returning an incorrect result, if it's given a malformed move. I didn't think through all the possible scenarios when I wrote it because I expected `is_move_malformed?` to always be called first.

The other consideration is that it potentially checks a large number of squares, which makes it the most computationally expensive step in the process. That means that it's best to call it last (i.e. only if needed).

You can do this as follows using the new pattern you just suggested:

```
def valid_move?(new_rank, new_file)
  return false if is_move_malformed?(new_rank, new_file)
  return false if (new_rank - rank).abs > 1 || (new_file - file).abs > 1
  !is_move_obstructed?(new_rank, new_file)
end
```

Or you can use the original pattern:

```
if !is_move_malformed?(new_rank, new_file) && (rank_diff <= 1) && (file_diff <= 1) && !is_move_obstructed?(new_rank, new_file) 
  true          
else
  false
end
```

The key is that when you chain boolean expressions using `&&` or `||`, Ruby will start evaluating from left to right and will go as far as needed to determine the result. That means that if you have three expressions chained with `&&` and the first one evaluates to `false`, Ruby will immediately return `false` for the full chain without evaluating expressions two or three.

brandoncha: oh...i understand now. it is a great point that i didn't catch it.
I will modify it. Thanks.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe King, type: :model do
+	let(:king) {create :king}
```
obulkin: Due to the associations between these models, creating an instance of `King` will also create a game, and that will in turn trigger the creation of 32 pieces via `populate_board!`, making the original king redundant. Instead, I would set up each test by creating a game and then selecting one of the kings associated with that game via `find_by`.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  describe "King's valid_move" do
+  	it "should be inside destination, blocked" do
+  		expect(king.valid_move?(1, 6)).to eq(false)
+  		expect(king.valid_move?(1, 4)).to eq(false)
+  		expect(king.valid_move?(2, 5)).to eq(false)
+  	end
```
obulkin: Could you add some more `expect` statements here to test diagonal moves and vertical moves down to a lower rank? You can use the black king for some of these.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it "shouldn't be outside destination" do
+      expect(king.valid_move?(1, 8)).to eq(false)
+      expect(king.valid_move?(1, 2)).to eq(false)
+      expect(king.valid_move?(3, 8)).to eq(false)
+    end
```
obulkin, line 67: The first two expectations here won't be met just because the move is obstructed, so if the goal with this block is to test the custom logic outside of `is_move_obstructed?` and `is_move_malformed?`, you should first make sure that the board and the move are set up in such a way that both of those will return `false`.

Also, as with the other blocks, could you please make sure to test all 8 possible move directions?

spec/models/king_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  	it "should be inside destination, not blocked" do
+  		game = create :game
+  		king2 = game.pieces.create(type: "King", rank: 4,  file: 5, color: "black")
+  		expect(king2.valid_move?(4, 6)).to eq(true)
+  		expect(king2.valid_move?(4, 4)).to eq(true)
+  		expect(king2.valid_move?(5, 6)).to eq(true)
+  	end
```
obulkin: As mentioned above, could you please test all 8 possible directions?

obulkin:
Please see my comments in line. The core code looks pretty good at this point, but the tests still need some work. 

Here is a chess visualization tool that I found pretty helpful when I was creating all the tests for `is_move_obstructed?`: http://www.apronus.com/chess/wbeditor.php. You might like it as well.

obulkin:
Also, after some more consideration, I think I was wrong to ask you to take out the test that involved malformed moves. We should probably include an example like this because it confirms that `valid_move?` is properly calling `is_move_malformed?`.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+
+RSpec.describe King, type: :model do
+  
+  describe "King's valid_move" do
+  	it "should be inside destination, blocked" do
+      game = create :game
```
obulkin: Since you're doing this as part of every test, you should make the code a bit more DRY by using `let` to create the `game` variable.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+
+RSpec.describe King, type: :model do
+  
+  describe "King's valid_move" do
+  	it "should be inside destination, blocked" do
```
obulkin: This description is pretty unclear. I would go with something along the lines of `"should properly evaluate blocked moves that are otherwise valid"`.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  		expect(king.valid_move?(1, 4)).to eq(false)
+  		expect(king.valid_move?(2, 5)).to eq(false)
+      expect(king.valid_move?(2, 4)).to eq(false)
+      expect(king.valid_move?(2, 6)).to eq(false)
+      expect(king2.valid_move?(7, 5)).to eq(false)
+      expect(king2.valid_move?(7, 4)).to eq(false)
+      expect(king2.valid_move?(7, 6)).to eq(false)
+  	end
+
+    it "shouldn't be outside destination" do
```
obulkin: Right here, my suggestion is `"should properly evaluate moves that are not allowed for a king but are otherwise valid"`.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      expect(king1.valid_move?(3, 4)).to eq(false)
+      expect(king1.valid_move?(3, 8)).to eq(false)
+      expect(king1.valid_move?(5, 4)).to eq(false)
+      expect(king1.valid_move?(5, 8)).to eq(false)
+      expect(king2.valid_move?(4, 3)).to eq(false)
+      expect(king2.valid_move?(4, 1)).to eq(false)
+      expect(king2.valid_move?(4, 5)).to eq(false)
+    end
+
+  	it "should be inside destination, not blocked" do
```
obulkin: Right here, I would go with something like `"should properly evaluate valid moves"`.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  		expect(king2.valid_move?(5, 6)).to eq(true)
+      expect(king2.valid_move?(3, 4)).to eq(true)
+      expect(king2.valid_move?(3, 6)).to eq(true)
+  	end
+
+    it "shouldn't be on same square after moving" do
+      game = create :game
+      king = game.pieces.find_by(rank: 1, file: 5, color: "white")
+      expect(king.valid_move?(1, 5)).to eq(false)
+    end
```
obulkin, line 48: Could you also add a test for moves that take a king outside the board but are otherwise valid? Since your version of `is_move_malformed?` doesn't handle that scenario, you can just comment those lines out for now, and we can quickly remove the comments later.

obulkin:
Could you add actual test logic to the last example and then deactivate it by turning the lines into comments? This way, we can just turn the test on once `is_move_malformed?` is ready by uncommenting the lines.

Otherwise, this looks ready to go. Thanks for cleaning up the formatting. It makes reviewing the code much more pleasant.

brandoncha:
I added 'a piece on the board?' logic in my version of `is_move_malformed?` to confirm that my tests are proper.

spec/models/king_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    it "should properly evaluate moves that take a king outside the board but are otherwise valid" do
+      king1 = game.pieces.create(type: "King", rank: 4, file: 8, color: "white")
+      king2 = game.pieces.create(type: "King", rank: 5, file: 1, color: "black")
+      expect(king1.valid_move?(4, 9)).to eq(false)
+      expect(king1.valid_move?(5, 9)).to eq(false)
+      expect(king1.valid_move?(3, 9)).to eq(false)
+      expect(king2.valid_move?(5, 0)).to eq(false)
+      expect(king2.valid_move?(6, 0)).to eq(false)
+      expect(king2.valid_move?(4, 0)).to eq(false)
+    end
```
obulkin, line 67: For completeness, could you also include scenarios where the new rank is outside the board?

obulkin:
Awesome! 👍  This looks good to go now. I'll merge.

PR #18, opened 2016-05-07
Updated is_move_malformed? method 

kenortiz:


app/models/piece.rb
```diff
@@ -44,57 +44,58 @@ def is_move_obstructed?(new_rank, new_file)
...
-        return true if piece_at_offset.present?         
+        next if exception_for_capture.call(offset, new_file - file, piece_at_offset)
+        return true if piece_at_offset.present?
       end
       false
     elsif new_file < file
       (new_file - file..-1).each do |offset|
         piece_at_offset = current_pieces.select{|piece| piece.rank == rank && piece.file == file + offset}
-        next if exception_for_capture.call(offset, new_file - file, piece_at_offset) 
-        return true if piece_at_offset.present?
```
kenortiz, line 86: I'm not exactly sure why the above code is showing as updated as I didn't modify it.

app/models/piece.rb
```diff
@@ -44,57 +44,58 @@ def is_move_obstructed?(new_rank, new_file)
...
   def is_move_malformed?(new_rank, new_file)
-      return true if rank == new_rank && file == new_file
-      return true if rank < 1 || rank > 8
-      return true if file < 1 || file > 8
-      return true if new_rank < 1 || new_rank > 8
-      return true if new_file < 1 || new_file > 8
+    return true if rank == new_rank && file == new_file
+    return true if new_rank < 1 || new_rank > 8
+    return true if new_file < 1 || new_file > 8
+    false
```
kenortiz, line 99: I removed the check for rank and file and added a return of false if the move is not malformed.

app/models/piece.rb
obulkin, line 86: Strange. It looks like some trailing whitespace characters may have gotten deleted for some reason. I'm not seeing any changes that would affect how the code works, so this shouldn't be an issue.

spec/models/piece_spec.rb
```diff
@@ -225,4 +225,25 @@
...
+      white_king = create :piece, rank: 1, file: 8, game: game
+      expect(white_king.send :is_move_malformed?, 0, 8).to eq(true)
+      expect(white_king.send :is_move_malformed?, 1, 9).to eq(true)
+      expect(white_king.send :is_move_malformed?, 8, 9).to eq(true)
+      expect(white_king.send :is_move_malformed?, 9, 8).to eq(true)
+      expect(white_king.send :is_move_malformed?, 8, 0).to eq(true)
+      expect(white_king.send :is_move_malformed?, 9, 1).to eq(true)
+      expect(white_king.send :is_move_malformed?, 1, 0).to eq(true)
+      expect(white_king.send :is_move_malformed?, 0, 1).to eq(true)
+    end
```
obulkin, line 257: Could you add some scenarios to this example that put both the rank and the file outside the board?

spec/models/piece_spec.rb
```diff
@@ -225,4 +225,25 @@
...
+  end
```
obulkin, line 258: In addition to the two examples provided, it would also be nice to have one that tests well-formed moves and makes sure that the method is returning `false` in those cases.

spec/models/piece_spec.rb
```diff
@@ -225,4 +225,25 @@
       end
     end
   end
+
+  describe "#is_move_malformed?" do
+    let(:game) {create :game}
```
obulkin: Creating a piece via the `:piece` factory will also lead to the creation of a game due to the association in the factory definition, so if you're going to be creating a piece at the start of every test, you can just do that via `let` instead: `let(:white_king) {create :piece}`.

kenortiz: Good point, thanks @obulkin!

spec/models/piece_spec.rb
```diff
@@ -244,6 +242,18 @@
...
+      expect(white_king.send :is_move_malformed?, 9, 0).to eq(true)
+      expect(white_king.send :is_move_malformed?, 0, 0).to eq(true)
+    end
+
+    it "should return false if a move is well-formed" do
+      expect(white_king.send :is_move_malformed?, 1, 2).to eq(false)
+      expect(white_king.send :is_move_malformed?, 2, 1).to eq(false)
+      expect(white_king.send :is_move_malformed?, 1, 8).to eq(false)
+      expect(white_king.send :is_move_malformed?, 8, 1).to eq(false)
+      expect(white_king.send :is_move_malformed?, 5, 4).to eq(false)
```
kenortiz, line 256: Added a test for well-formed moves.

spec/models/piece_spec.rb
```diff
@@ -244,6 +242,18 @@
       expect(white_king.send :is_move_malformed?, 9, 1).to eq(true)
       expect(white_king.send :is_move_malformed?, 1, 0).to eq(true)
       expect(white_king.send :is_move_malformed?, 0, 1).to eq(true)
+      expect(white_king.send :is_move_malformed?, 0, 9).to eq(true)
+      expect(white_king.send :is_move_malformed?, 9, 9).to eq(true)
+      expect(white_king.send :is_move_malformed?, 9, 0).to eq(true)
+      expect(white_king.send :is_move_malformed?, 0, 0).to eq(true)
```
kenortiz, line 248: Added scenarios which put moves outside the board.

obulkin:
👍  This looks good to go now. I'll merge it in.

PR #19, opened 2016-05-10
Added Pawn#valid_move? and the Piece#capture_move? helper

obulkin:
- Created a `capture_move?` helper on the piece model for determining whether a move would entail the capture of an enemy piece
- Wrote unit tests for `Piece#capture_move?`
- Added a `valid_move?` method to the pawn model
- Created a pawn factory that inherits from the basic piece factory
- Wrote unit tests for `Pawn#valid_move?`

### DIFF
--- a/app/models/pawn.rb
+++ b/app/models/pawn.rb
@@ -1,2 +1,22 @@
 class Pawn < Piece
+  def valid_move?(new_rank, new_file)
+    rank_diff = new_rank - rank
+    file_diff = new_file - file
+
+    return false if is_move_malformed?(new_rank, new_file) || is_move_obstructed?(new_rank, new_file)
+
+    if color == "white"
+      # Standard pawn move
+      return true if rank_diff == 1 && file_diff == 0
+      # Double move
+      return true if rank == 2 && rank_diff == 2 && file_diff == 0
+      # Capture move
+      return true if rank_diff == 1 && file_diff.abs == 1 && capture_move?(new_rank, new_file)
+    else
+      return true if rank_diff == -1 && file_diff == 0
+      return true if rank == 7 && rank_diff == -2 && file_diff == 0
+      return true if rank_diff == -1 && file_diff.abs == 1 && capture_move?(new_rank, new_file)
+    end
+    false
+  end
 end
--- a/app/models/pawn.rb
+++ b/app/models/pawn.rb
@@ -6,17 +6,19 @@ class Pawn < Piece
     return false if is_move_malformed?(new_rank, new_file) || is_move_obstructed?(new_rank, new_file)
 
     if color == "white"
-      # Standard pawn move
-      return true if rank_diff == 1 && file_diff == 0
-      # Double move
-      return true if rank == 2 && rank_diff == 2 && file_diff == 0
-      # Capture move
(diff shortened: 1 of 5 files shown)


app/models/pawn.rb
```diff
@@ -1,2 +1,22 @@
 class Pawn < Piece
-end
\ No newline at end of file
+  def valid_move?(new_rank, new_file)
+    rank_diff = new_rank - rank
+    file_diff = new_file - file
+
+    return false if is_move_malformed?(new_rank, new_file) || is_move_obstructed?(new_rank, new_file)
```
lasley, line 6: Might make sense to push this into the `Piece` model, then call `super` here

app/models/piece.rb
```diff
@@ -11,6 +11,15 @@ def user
     color == "white" ? game.white_player : game.black_player
   end
 
+  # Please note that this method does not test whether the move is valid
+  def capture_move?(new_rank, new_file)
+    if color == "white"
+      game.pieces.find_by(rank: new_rank, file: new_file, color: "black") ? true : false
+    else
+      game.pieces.find_by(rank: new_rank, file: new_file, color: "white") ? true : false
```
lasley: I would assign a color variable in the if, then unify these two queries using the `color` var instead. This is good too, just duplicating some important code.

app/models/pawn.rb
```diff
@@ -1,2 +1,22 @@
...
+      # Standard pawn move
+      return true if rank_diff == 1 && file_diff == 0
+      # Double move
+      return true if rank == 2 && rank_diff == 2 && file_diff == 0
+      # Capture move
+      return true if rank_diff == 1 && file_diff.abs == 1 && capture_move?(new_rank, new_file)
+    else
+      return true if rank_diff == -1 && file_diff == 0
+      return true if rank == 7 && rank_diff == -2 && file_diff == 0
+      return true if rank_diff == -1 && file_diff.abs == 1 && capture_move?(new_rank, new_file)
```
lasley: Little math trick that may help dedup a bit here, you can set a modifier variable then multiply against that to create your positive negatives. X \* 1 is always X, and X \* -1 is always -X. 

With that and a var for your rank position, all of the returns could then be unified with those.

app/models/pawn.rb
obulkin, line 6: I considered suggesting that in the context of `King#valid_move?`, but at the time, it seemed like the ideal flow from an efficiency standpoint would be the following, prohibiting the use of `super`:
- Call `is_move_malformed?`
- Execute type-specific logic
- Call `is_move_obstructed?`

Hilariously, I immediately violated that pattern here because the pawn logic is fairly complex and writing it out in terms of valid moves rather than failure conditions made the code much more clear. However, I was thinking we could return to it in the other versions of `valid_move?`.

lasley, line 6: Hah ok this makes sense. I'll yield to the design decision here

app/models/piece.rb
obulkin: Makes sense. I'll do that.

app/models/pawn.rb
obulkin: I see. OK. I'll play around with that.

lasley:
👍

kenortiz:
This looks good to me. Merging.

PR #22, opened 2016-05-13
valid_move? for Rook

brandoncha:
Please review and let me know if there are some issue to modify.


app/models/piece.rb
```diff
@@ -104,4 +104,13 @@ def is_move_malformed?(new_rank, new_file)
...
     false
   end
+
+  def horizontal_move?(new_rank, new_file)
+    rank == new_rank
+  end
+
+  def vertical_move?(new_rank, new_file)
+    file == new_file
+  end
```
obulkin: I think it's overkill to write a separate method for a single boolean comparison. The code would be easier to follow (no need to look at two separate files), more concise, and marginally more efficient if you just did this inline as part of `Rook#valid_move?`.

spec/models/rook_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      expect(white_rook.valid_move?(0, 2)).to eq(false)
+      expect(white_rook.valid_move?(2, 0)).to eq(false)
+      expect(white_rook.valid_move?(1, 1)).to eq(false)
+      expect(black_rook.valid_move?(8, 1)).to eq(false)
+      expect(black_rook.valid_move?(7, 0)).to eq(false)
+      expect(black_rook.valid_move?(8, 0)).to eq(false)
+      expect(black_rook.valid_move?(9, 0)).to eq(false)
+      expect(black_rook.valid_move?(9, 1)).to eq(false)
+      expect(black_rook.valid_move?(9, 2)).to eq(false)
+    end
```
obulkin, line 26: Could you please cover the following cases here as well?
- A rank that is too large and a file that is too large
- A rank that is too small and a file that is too large
- A valid rank and a file that is too large

spec/models/rook_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    it "should properly evaluate obstructed moves that are otherwise valid" do
+      white_rook = game.pieces.find_by(rank: 1, file: 1)
+      black_rook = game.pieces.find_by(rank: 8, file: 1)
+      expect(white_rook.valid_move?(1, 2)).to eq(false)
+      expect(white_rook.valid_move?(2, 1)).to eq(false)
+      expect(black_rook.valid_move?(8, 2)).to eq(false)
+      expect(black_rook.valid_move?(7, 1)).to eq(false)
+    end
```
obulkin, line 37: Right here, could you also cover an obstructed horizontal move over to a lower file? I would take care of this by grabbing the black rook at h8, instead of the one at a8.

spec/models/rook_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    it "should properly evaluate unobstructed moves that are not allowed for a Rook but are otherwise valid" do
+      white_rook = game.pieces.create(type: "Rook", rank: 4, file: 4, color: "white")
+      expect(white_rook.valid_move?(3, 5)).to eq(false)
+      expect(white_rook.valid_move?(5, 5)).to eq(false)
+      expect(white_rook.valid_move?(3, 3)).to eq(false)
+      expect(white_rook.valid_move?(3, 6)).to eq(false)
+    end
```
obulkin, line 49: Optional: It might be nice to test the fourth diagonal and a few more L shaped moves here but feel free to skip that if you think it's excessive.

spec/models/rook_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      expect(black_rook.valid_move?(7, 0)).to eq(false)
+      expect(black_rook.valid_move?(8, 0)).to eq(false)
+      expect(black_rook.valid_move?(9, 0)).to eq(false)
+      expect(black_rook.valid_move?(9, 1)).to eq(false)
+      expect(black_rook.valid_move?(9, 2)).to eq(false)
+      expect(black_rook.valid_move?(8, 1)).to eq(false)
+      expect(black_rook.valid_move?(15, -9)).to eq(false)
+      expect(black_rook.valid_move?(9, -9)).to eq(false)
+      expect(black_rook.valid_move?(8, -8)).to eq(false)
+    end
```
obulkin, line 26: The `expect` statements that you added here don't actually cover the scenarios that I mentioned, all of which involve files greater than 8. Here they are again for reference:
- `rank > 8 && file > 8`
- `rank < 1 && file > 8`
- A valid rank and `file > 8`

Basically, I'd like to cover the eight directions in which you can go outside the board (two vertical, two horizontal, and four diagonal).

obulkin:
Please see my comment in line. Also, it looks like you forgot to delete the `horizontal_move?` and `vertical_move?` helpers.

spec/models/rook_spec.rb
brandoncha, line 26: I see. I will create two more Rooks(rank: 8, file: 8)&(rank:1, file:8) and test the eight direction scenarios.

app/models/rook.rb
```diff
@@ -1,2 +1,11 @@
 class Rook < Piece
-end
\ No newline at end of file
+    def valid_move?(new_rank, new_file)
+      if !is_move_malformed?(new_rank, new_file) && (rank == new_rank || file == new_file) && !is_move_obstructed?(new_rank, new_file)
+        true
+      else
+        false
+    end
+  end
```
obulkin, line 8: There's an indentation issue here that makes the code a bit confusing at first sight. The `end` on line 8 should align with the `def` on line 2, while the `end` on line 7 should align with the `if`and `else` lines.

brandoncha, line 8: yes, i should check this. will modify it.

obulkin:
Nice. 👍  This looks good to go now, so I'll merge.

PR #25, opened 2016-05-19
Board

aryawinterfell:
This is currently where I stand with getting the pieces on the board. Going to try and figure out how to get the images on the board now... let me know your thoughts. Thanks


app/views/boards/show.html.erb
```diff
@@ -1,15 +1,23 @@
 <body>
  <div class="container">
  <table class="block-center">
-     <% 4.times do %>
+     <% 4.times do |row| %>
       <tr class="even">
- 	     <% 8.times do %>
- 	      <td></td>	 
+ 	     <% 8.times do |even_cell| %>
+       <% if Piece.where(file: even_cell + 1, rank: row * 2 + 1)[0]%>
```
obulkin: There are two problems with this approach:
- It's looking at all of the pieces in the DB, rather than the pieces associated with a specific game
- In the worst case scenario (a full board), you'll wind up running 64 `where` queries to check each cell and 32 more to get the type associated with each piece, which is very inefficient

The solution is as follows:
- Get rid of the temporary controller/route that you created earlier and make this the view for `GamesController#show`, which will give you access to a game variable (`@game`)
- In the controller, get all of the pieces associated with the game and assign them to an instance variable (e.g. `@game_pieces`). Then, in the view, use your instance variable together with `select` (http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-select) to check whether a piece is present in a specific place and retreive that piece's type. This way, there will be one initial DB query and a bunch of in-memory operations, making this process much more efficient.

aryawinterfell: Ya, I totally get what you're saying. I originally had  it where I used the game_id - 

<% if Piece.where(game_id: 1, file: even_cell + 1, rank: row \* 2 + 1)[0]%> etc...  
game_id 1 was the fake game i created 

And in the controller I had 

def show
    @my_game = Game.find(1)
  end

Trying to figure out if I can somehow use the game_id to pull the id of the game... hmm But, I'm looking at what you said too. maybe that's what you were saying

obulkin: `GamesController#show` is already pulling up the right game based on `params[:id]`, so you can just use the Active Record association between games and pieces to get the proper set of pieces.

app/models/bishop.rb
```diff
@@ -10,4 +10,12 @@ def valid_move?(new_rank, new_file)
     
     !is_move_obstructed? new_rank, new_file
   end
+
+  def image
+    if color == 'white'
+      return '&#9815;'
+    else
+      return '&#9821;'
```
lasley, line 18: I like the use of Unicodes here. Hint - the black is +6 from white, so something like this could reduce repetition:

```
mod = color == 'white' ? 0 : 6
return "&##{9815 + mod}"
```

app/views/games/show.html.erb
```diff
@@ -0,0 +1,26 @@
+<body>
+ <div class="container">
+ <table class="block-center">
+     <% 4.times do |row| %>
+      <tr class="even">
+       <% 8.times do |even_cell| %>
+       <% if @pieces.where(file: even_cell + 1, rank: row * 2 + 1)[0]%>
+        <td><font class="unicode"><%= @pieces.where(file: even_cell + 1, rank: row * 2 + 1)[0].image.html_safe %></font></td>
```
lasley: Stylistic, but IMO it looks much better if each HTML element creates a visual nesting w/ indentation similar to Ruby def/end blocks. Makes it easier to spot what's going on.

```
<td>
    <font class="unicode">
        <%= ...
```

lasley:
LGTM no blockers 👍

app/models/bishop.rb
obulkin, line 18: Optional: My preferred approach would probably be `color == "white" ? "&#9815;" : "&#9821;"` because it neatly reduces everything down to one line. So many options! 😄

app/views/games/show.html.erb
```diff
@@ -0,0 +1,26 @@
...
+       <% if @pieces.where(file: even_cell + 1, rank: row * 2 + 1)[0]%>
+        <td><font class="unicode"><%= @pieces.where(file: even_cell + 1, rank: row * 2 + 1)[0].image.html_safe %></font></td> 
+        <% else %>
+        <td></td> 
+        <% end %>   
+       <% end %>
+      </tr>
+      <tr class="odd">
+       <% 8.times do |odd_cell| %>
+        <% if @pieces.where(file: odd_cell + 1, rank: row * 2 + 2)[0]%>
```
obulkin, line 20: You're still using the `where` method, which will trigger all those DB queries I mentioned. Now that you have all the pieces associated with the game stored in memory as `@pieces`, the way to avoid that is to use the `select` method instead: http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-select

obulkin:
You're getting closer, but you haven't resolved the efficiency issue yet. Please see my comment in line for more info.

Also, there are a number of additional files associated with the temporary controller that still need to be removed:
- `app/assets/javascript/boards.js.coffee`
- `app/assets/stylesheets/boards.scss` (the CSS rules here should move to `games.scss`)
- `app/helpers/boards_helper.rb`

obulkin:
Since the team voted to include this branch in the first release of the app as is, I'm merging it without further review. The efficiency issue mentioned above has not been resolved yet.

PR #26, opened 2016-05-21
Added valid_move? for queen and added unit tests

kenortiz:
Please review and let me know if any changes are needed.
- Added valid_move? to Queen model
- Added unit test for Queen#valid_move?

### DIFF
--- a/app/models/queen.rb
+++ b/app/models/queen.rb
@@ -1,2 +1,7 @@
 class Queen < Piece
+  def valid_move?(new_rank, new_file)
+    return false if is_move_malformed?(new_rank, new_file) || is_move_obstructed?(new_rank, new_file)
+    
+    rank == new_rank || file == new_file || (rank - new_rank).abs == (file - new_file).abs
+  end
 end
--- a/spec/models/queen_spec.rb
+++ b/spec/models/queen_spec.rb
@@ -1,0 +1,92 @@
+require 'rails_helper'
+
+RSpec.describe Queen, type: :model do
+  describe "#valid_move?" do
+    let(:game) {create :game}
+
+    it "should properly evaluate malformed moves that would otherwise be valid" do
+      white_queen = game.pieces.find_by(rank: 1, file: 1)
+      white_queen2 = game.pieces.find_by(rank: 1, file: 8)
+      white_queen3 = game.pieces.find_by(rank: 1, file: 3)
+      black_queen = game.pieces.find_by(rank: 8, file: 1)
+      black_queen2 = game.pieces.find_by(rank: 8, file: 8)
+      black_queen3 = game.pieces.find_by(rank: 8, file: 3)
+      expect(white_queen.valid_move?(1, 1)).to eq(false)
+      expect(white_queen.valid_move?(0, 1)).to eq(false)
+      expect(white_queen.valid_move?(1, 0)).to eq(false)
+      expect(white_queen.valid_move?(0, 0)).to eq(false)
+      expect(white_queen2.valid_move?(1, 9)).to eq(false)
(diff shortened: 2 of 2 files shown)


app/models/queen.rb
```diff
@@ -1,2 +1,7 @@
 class Queen < Piece
-end
\ No newline at end of file
+  def valid_move?(new_rank, new_file)
+    return false if is_move_malformed?(new_rank, new_file) || is_move_obstructed?(new_rank, new_file)
+    
+    rank == new_rank || file == new_file || (rank - new_rank).abs == (file - new_file).abs
```
obulkin, line 5: I would suggest reworking this slightly so that `is_move_obstructed?` only gets called at the end if needed. It's the most expensive step in the process (it needs to hit the DB and will usually loop a number of times), so that should make `Queen#valid_move?` more efficient.

spec/models/queen_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  describe "#valid_move?" do
+    let(:game) {create :game}
+
+    it "should properly evaluate malformed moves that would otherwise be valid" do
+      white_queen = game.pieces.find_by(rank: 1, file: 1)
+      white_queen2 = game.pieces.find_by(rank: 1, file: 8)
+      white_queen3 = game.pieces.find_by(rank: 1, file: 3)
+      black_queen = game.pieces.find_by(rank: 8, file: 1)
+      black_queen2 = game.pieces.find_by(rank: 8, file: 8)
+      black_queen3 = game.pieces.find_by(rank: 8, file: 3)
```
obulkin, line 13: The pieces that you're pulling up here via `find_by` won't actually be queens, which means that you're testing other copies of `valid_move?`, rather than `Queen#valid_move?`

spec/models/queen_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      expect(white_queen.valid_move?(5, 2)).to eq(false)
+      expect(white_queen.valid_move?(5, 6)).to eq(false)
+      expect(white_queen2.valid_move?(5, 7)).to eq(false)
+      expect(white_queen2.valid_move?(5, 3)).to eq(false)
+      expect(white_queen2.valid_move?(6, 6)).to eq(false)
+      expect(white_queen2.valid_move?(6, 4)).to eq(false)
+      expect(white_queen2.valid_move?(3, 7)).to eq(false)
+      expect(white_queen2.valid_move?(3, 3)).to eq(false)
+      expect(white_queen2.valid_move?(2, 6)).to eq(false)
+      expect(white_queen2.valid_move?(2, 4)).to eq(false)
```
obulkin, line 69: The last two moves here are obstructed.

spec/models/queen_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      expect(white_queen.valid_move?(2, 4)).to eq(true)
+      expect(white_queen.valid_move?(5, 4)).to eq(true)
+      expect(white_queen.valid_move?(6, 4)).to eq(true)
+      expect(white_queen2.valid_move?(5, 4)).to eq(true)
+      expect(white_queen2.valid_move?(5, 6)).to eq(true)
+      expect(white_queen2.valid_move?(7, 4)).to eq(true)
+      expect(white_queen2.valid_move?(7, 6)).to eq(true)
+      expect(white_queen2.valid_move?(3, 2)).to eq(true)
+      expect(white_queen2.valid_move?(4, 7)).to eq(true)
+    end
```
obulkin, line 90: Could you add a couple cases here that cover diagonal moves with increasing rank?

lasley:
Nothing to add here other than what Oleg noted. Looks good though!

PR #27, opened 2016-05-21
Added environment variable for production secret token.

kenortiz:
Please review and let me know if any changes are needed.

I will update production with secret token once approved.

### DIFF
--- a/config/initializers/secret_token.rb
+++ b/config/initializers/secret_token.rb
@@ -9,4 +9,8 @@
 
 # Make sure your secret_key_base is kept private
 # if you're sharing your code publicly.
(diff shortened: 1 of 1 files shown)


config/initializers/secret_token.rb
```diff
@@ -9,4 +9,8 @@
 
 # Make sure your secret_key_base is kept private
 # if you're sharing your code publicly.
-MANIACChess::Application.config.secret_key_base = 'f4e0e54e5ad27a1e9a73f9be70801966b90880e25bd75cce40483789d439deeec6d1224d24fdc208d907640f5dddd6013cff2c62204250dbb8480fdcda375def'
+MANIACChess::Application.config.secret_key_base = if Rails.env.development? or Rails.env.test?
```
obulkin: Even though it should work, I think the use of `or` here is probably a bad practice, and `||` would be better. `or` and `and` are intended for use in flow control, not boolean expressions, and have different precedence than `||` and `&&`, which can lead to unexpected results when they are used interchangeably. See here for more info: http://devblog.avdi.org/2010/08/02/using-and-and-or-in-ruby/

